PR #41, opened 2017-06-08
New shortName field added to Section

tenochdlc:
New shortName field added to Section, modified Summary to use this if available, normal title if not

The gform microservice was also modified, branch with the same name.


tenochdlc:
Tests added

PR #54, opened 2017-06-29
version is case class

pawelpanasewicz:


lutnos:
I think we probably ought ti have a discussion to agree a Camel-Case convention for the name gform, at least.

I do not know whether the organisation has a name spelling and capitalisation standard

PR #105, opened 2017-07-25
EEiTT-866 - Initial work to support repeating sections

tenochdlc:
I am trying to avoid a massive PR
This is some initial work to support repeating sections on the frontend

### DIFF
--- a/test/uk/gov/hmrc/gform/services/NumberValidationSpec.scala
+++ b/test/uk/gov/hmrc/gform/services/NumberValidationSpec.scala
@@ -16,21 +16,19 @@
 
 package uk.gov.hmrc.gform.services
 
-import java.time.LocalDate
-
 import cats.scalatest.EitherMatchers
-import cats.scalatest.ValidatedValues._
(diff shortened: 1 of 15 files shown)


test/uk/gov/hmrc/gform/services/NumberValidationSpec.scala
```diff
@@ -224,5 +222,4 @@ class NumberValidationSpec extends FlatSpec with Matchers with EitherMatchers wi
     result.toEither should beLeft(Map(fieldValue.id -> Set("number must be at most 2 whole digits and decimal fraction must be at most 1 digits")))
   }
 
-  implicit lazy val hc = HeaderCarrier()
```
pawelpanasewicz: `hc` was placed at the end of the file intentionally because as we agreed "important goes up". Here test is more important and should be seen first.

pawelpanasewicz:
LGTM. If you could just bring `hc`s back to bottom or if there are some special intentions behind keeping them up front - rebase and merge :)

PR #106, opened 2017-07-26
changing delete to a get endpoint to work with current functionality

DanielConnelly:


app/uk/gov/hmrc/gform/views/hardcoded/pages/confirm_delete.scala.html
```diff
@@ -7,7 +7,7 @@
 
 @uk.gov.hmrc.gform.views.html.main_template(title = "Confirm Delete", bodyClasses = None) {
 
-<form action="@uk.gov.hmrc.gform.controllers.routes.FormController.delete(formTypeId, formId)" method="POST">
+<form action="@uk.gov.hmrc.gform.controllers.routes.FormController.delete(formTypeId, formId)" method="GET">
```
pawelpanasewicz: better would be DELETE

DanielConnelly: https://softwareengineering.stackexchange.com/questions/114156/why-are-there-are-no-put-and-delete-methods-on-html-forms DELETE can not be supported in the form in HTML5

PR #107, opened 2017-07-26
Eeitt 730

DanielConnelly:


app/uk/gov/hmrc/gform/auth/Auth.scala
```diff
@@ -0,0 +1,46 @@
...
+import uk.gov.hmrc.gform.models.UserId
+import uk.gov.hmrc.gform.models.userdetails.AffinityGroup
+import uk.gov.hmrc.gform.models.userdetails.AffinityGroup._
+import uk.gov.hmrc.play.frontend.auth.AuthContext
+import uk.gov.hmrc.play.frontend.auth.connectors.AuthConnector
+import uk.gov.hmrc.play.http.HeaderCarrier
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+class Auth @Inject() (
```
pawelpanasewicz: Looks like it is kind of Service. Maybe it's worth to rename it to `AuthService` or even better to `AuthXXXService` (replace XXX with meaningful word). Since it is a service its methods should return objects different than `play.mvc.Result`. Such conversion should happen only in `Controller` layer.

app/uk/gov/hmrc/gform/auth/Auth.scala
```diff
@@ -0,0 +1,46 @@
...
+    eeittAuth: EeittAuth
+) {
+
+  def doAuth(formTemplate: FormTemplate, authCase: UserId => Future[Result])(implicit authContext: AuthContext, hc: HeaderCarrier, ex: ExecutionContext): Future[Result] = {
```
pawelpanasewicz: doAuthenticate or doAuthorise?

app/uk/gov/hmrc/gform/auth/Auth.scala
```diff
@@ -0,0 +1,46 @@
...
+    formTemplate.authConfig.authModule.value match {
+      case "legacyEEITTAuth" => eeittAuth.legacyAuth(formTemplate, authCase)
+      case _ => Future.successful(Ok) //TODO Dave New Auth Method
```
pawelpanasewicz: Returning ok is not the best option. Rather we should return `Unauthorised`. We don't want to allow to get access is there is missing config ;)

app/uk/gov/hmrc/gform/auth/Auth.scala
pawelpanasewicz: We don't need @Inject here. Object is created manually in AuthModule

app/uk/gov/hmrc/gform/auth/EeittAuth.scala
```diff
@@ -0,0 +1,74 @@
...
+        x.asOpt[AffinityGroup] match {
+          case Some(affinGroup) => JsSuccess(UserDetails(userId, affinGroup))
+          case None => JsError("No AffinityGroup is present")
+        }
+      case None => JsError("No UserId is present")
+    }
+  }
+}
+
+class EeittAuth @Inject() (
```
pawelpanasewicz: Let's remove `@Inject` and and wire object manually. In future we are going to remove injector from project and rely on manually wiring which is safer.

app/uk/gov/hmrc/gform/auth/EeittAuth.scala
```diff
@@ -0,0 +1,74 @@
...
+    eeittConnector: EeittConnector,
+    authConnector: AuthConnector,
+    eeittUrl: String,
+    baseUrl: String
+) {
+
+  def legacyAuth(formTemplate: FormTemplate, action: UserId => Future[Result])(implicit authContext: AuthContext, hc: HeaderCarrier, ex: ExecutionContext): Future[Result] =
```
pawelpanasewicz: `ExecutionContext` can be part of constructor list. After that all methods in service can use this value and get rid of additional paramete which is common to all methods.

pawelpanasewicz:
Let's talk about it :)

app/uk/gov/hmrc/gform/auth/AuthModule.scala
```diff
@@ -36,11 +38,22 @@ class AuthModule @Inject() (configModule: ConfigModule, wSHttpModule: WSHttpModu
...
+  lazy val authorisationService: AuthorisationService = new AuthorisationService(eeittAuthorisationDelegate, authConnector)
+
+  lazy val authenticatedBy: auth.Actions#AuthenticatedBy = new authActions.AuthenticatedBy(governmentGateway, taxRegime, alwaysVisiblePageVisibility)
+
+  lazy val eeittConnector = new EeittConnector(
+    s"${configModule.serviceConfig.baseUrl("eeitt")}/eeitt",
+    wSHttpModule.auditableWSHttp
+  )
+
+  def redirectToEeitt(formTypeId: FormTypeId): Future[Result] =
```
pawelpanasewicz: Module suppose to have only instances not functions. Lets move it to proper controller.

app/uk/gov/hmrc/gform/auth/EeittAuthorisationDelegate.scala
```diff
@@ -0,0 +1,53 @@
...
+    eeittConnector: EeittConnector,
+    appConfig: AppConfig
+) {
+
+  def legacyAuth(formTemplate: FormTemplate, userDetails: UserDetails)(implicit hc: HeaderCarrier, ex: ExecutionContext): Future[AuthResult] =
+    for {
+      isOk <- eeittConnector.isAllowed(userDetails.userId.value, formTemplate.authConfig.regimeId, userDetails.affinityGroup)
+    } yield isAuthed(isOk, formTemplate.formTypeId, userDetails.userId)
+
+  private def isAuthed(isOk: Verification, formTypeId: FormTypeId, userId: UserId): AuthResult = {
```
pawelpanasewicz: formTypeId and userId aren't use here :)

app/uk/gov/hmrc/gform/service/PrepopService.scala
```diff
@@ -41,7 +41,7 @@ object AuthContextPrepop {
 
 object PrepopService {
 
-  def eeittConnector = EeittConnector // TODO - DI this
+  def eeittConnector: EeittConnector = ??? // TODO - DI this
```
pawelpanasewicz: Look like it's not used

PR #123, opened 2017-08-04
Eeitt 866 - PR 3/3 Repeating sections

tenochdlc:


pawelpanasewicz:
There is ExampleData trait/object in upcomming commit which provides bunch of reusable data. I encourage to use it.

PR #126, opened 2017-08-07
adding in functionality for money formats and creating the sort code …

DanielConnelly:
…flow


lutnos:
It would be nice to have the story id, EEITT-858, it a branch name or commit message somewhere.

This needs a little more work, sorry.

The validation checks for the account number need to validate 8 digits and not less; similarly, that each part of the sort code is two digits.

Thanks

PR #130, opened 2017-08-08
adding in contact details format for the forms

DanielConnelly:


test/uk/gov/hmrc/gform/services/FormatValidationSpec.scala
```diff
@@ -166,4 +166,84 @@ class FormatValidationSpec extends Spec {
     result.toEither should beLeft(Map(fieldValue.id -> Set("must be a whole number")))
   }
 
+  "TelephoneNumber" should "be valid within limit of 30" in {
+    val textConstrait = TelephoneNumber
```
pawelpanasewicz: How about sharing this fixtures in `ExampleData` and using it in this test?

lutnos:
Nathan has sold you a dodgy regex, I believe that the aA-Zz should be A-Za-z .  Do you see why?

Also I think he said telephone numbers were text 4,30, not text 0,30

app/uk/gov/hmrc/gform/validation/ValidationService.scala
```diff
@@ -192,51 +191,66 @@ class ComponentsValidator(fieldValue: FieldValue, data: Map[FieldId, Seq[String]
...
+      case (_, value :: Nil, TelephoneNumber) => textValidator(value, 0, 30, true)
+      case (_, value :: Nil, Email) => email(value)
       case (_, value :: Nil, Number(maxWhole, maxFractional, _)) => validateNumber(value, maxWhole, maxFractional, false)
       case (_, value :: Nil, PositiveNumber(maxWhole, maxFractional, _)) => validateNumber(value, maxWhole, maxFractional, true)
       case (_, value :: rest, _) => Valid(()) // we don't support multiple values yet
     }
   }
 
-  private def checkLength(value: String, desiredLength: Int) = {
+  private def textValidator(value: String, min: Int, max: Int, mandatory: Boolean) =
```
tenochdlc: Shouldn't we also be checking for valid chars? 0-9, "ext", space, ".", "+","(", ")"

lutnos: Yes, but I think that should probably be done as part of https://jira.tools.tax.service.gov.uk/browse/EEITT-848

PR #133, opened 2017-08-10
Eeitt 848

DanielConnelly:


app/uk/gov/hmrc/gform/validation/ValidationService.scala
```diff
@@ -212,7 +215,17 @@ class ComponentsValidator(fieldValue: FieldValue, data: Map[FieldId, Seq[String]
     value match {
       case UTR() => Valid(())
       case x if Nino.isValid(x) => Valid(())
-      case _ => Invalid(Map(fieldValue.id -> Set(fieldValue.errorMessage.getOrElse("Not a valid Id"))))
+      case _ => Invalid(Map(fieldValue.id -> errors("Not a valid Id")))
+    }
+  }
+
+  def basicValidation(value: String, constraint: TextConstraint) = {
```
pawelpanasewicz: Now it's hard to guess what caused validation error. How about redesigning it in such way that user gets error message telling the cause of not accepting his input?

2. The wildcard in the function will accept everything.  If someone adds more cases it can be messy and compiler won't help you when not matching an specific case. Lets talk.

pawelpanasewicz:
well done :+1:

PR #142, opened 2017-08-14
Daniel/sort code

DanielConnelly:


lutnos:
See https://github.com/hmrc/gform/pull/106

Here and in front end, should iterate over the field IDs of a sequence.  Also, is it safe to use starts-with?  Later ids may also start with that, but not followed by a hyphen.  In general it would be safer and clearer to generate the Ids you are working with by suffixing the known one

app/uk/gov/hmrc/gform/models/helpers/Fields.scala
```diff
@@ -30,7 +30,7 @@ object Fields {
   def okValues(formFieldMap: Map[FieldId, Seq[String]], fieldValues: List[FieldValue], repeatService: RepeatingComponentService, envelope: Envelope)(fieldValue: FieldValue)(implicit hc: HeaderCarrier): Option[FormFieldValidationResult] = {
     val formFields = toFormField(formFieldMap, fieldValues, repeatService).map(hf => hf.id -> hf).toMap
     fieldValue.`type` match {
-      case Address(_) | Date(_, _, _) =>
+      case Address(_) | Date(_, _, _) | UkSortCode(_) =>
         val fieldOkData =
           formFields.filter {
             case (fieldId, formField) => fieldId.value.startsWith(fieldValue.id.value) // Get just fieldIds related to fieldValue
```
lutnos: Surely code like this will fail when an address, date or sort code field id is a prefix of another one.  For example a field with an id of "1" and another with an id of "12".  Attempts to process id "1"  will also get data from fields with ID "12".  While you could start matching for the suffix character too in the filter, I think that it would be safer and clearer to generate the ids that you need for each field type, for example x_1 x_2 and x_3 for a sort code, x_street1, x_street2 ... for address.  Do you see?

PR #164, opened 2017-08-17
exposed api for deleting uploaded file

pawelpanasewicz:


app/uk/gov/hmrc/gform/restapi/RestApiController.scala
```diff
@@ -14,23 +14,36 @@
  * limitations under the License.
  */
 
-package uk.gov.hmrc.gform.config
+package uk.gov.hmrc.gform.restapi
```
tenochdlc: this is for fileupload at the moment, so I think the package name should reflect that

app/uk/gov/hmrc/gform/restapi/RestApiController.scala
```diff
@@ -14,23 +14,36 @@
...
+  controllersModule: ControllersModule
+)
+    extends FrontendController {
+  import uk.gov.hmrc.gform.controllers.AuthenticatedRequest._
 
-class ConfigController @Inject() (gformBackendModule: GformBackendModule) extends FrontendController {
+  //TODO: use proxy actions here
+  //TODO: proxy actions must rely on WsHttp
 
   def exposedConfig() = Action.async { implicit r =>
```
tenochdlc: Why do you need to go all the way to the backend to retrieve information that doesn't change at run time. Why not in a config file? and an IT test to make sure the backend and JS match

app/uk/gov/hmrc/gform/restapi/RestApiController.scala
```diff
@@ -14,23 +14,36 @@
...
+class RestApiController @Inject() (
+  gformBackendModule: GformBackendModule,
+  controllersModule: ControllersModule
+)
+    extends FrontendController {
+  import uk.gov.hmrc.gform.controllers.AuthenticatedRequest._
 
-class ConfigController @Inject() (gformBackendModule: GformBackendModule) extends FrontendController {
+  //TODO: use proxy actions here
+  //TODO: proxy actions must rely on WsHttp
```
tenochdlc: Probably better to provide more background on this comment, if some one else were to pick up this TODO I don't think they would have enough details.

app/uk/gov/hmrc/gform/restapi/RestApiController.scala
pawelpanasewicz: Idea was to have common controller for all functionality which is only in backend. At the moment we expose api for: 
1. exposed-config
2. delete-file 

Note that controller does simply calling `GformConnector` (no logic here at all). This is why I wanted to group such calls in one group (package/controller). Later here would land all other functionalities exposed for 3rd party developers.

I hope this brings some more light into it.

app/uk/gov/hmrc/gform/restapi/RestApiController.scala
pawelpanasewicz: Exposed config is indeed read from config file which lays in `GformBackend`. It's there because GformBackend uses this config as well.

conf/app.routes
```diff
@@ -17,4 +17,5 @@ POST        /summary/:formId/:formTemplateId                             @uk.gov
 
 GET         /assets/*file                                                controllers.Assets.versioned(path="/public", file: Asset)
 
-GET         /api/exposed-config                                          @uk.gov.hmrc.gform.config.ConfigController.exposedConfig()
+GET         /api/exposed-config                                          @uk.gov.hmrc.gform.restapi.RestApiController.exposedConfig()
```
pawelpanasewicz: I don't understand what's wrong in here?

app/uk/gov/hmrc/gform/restapi/RestApiController.scala
pawelpanasewicz: Sure, I'll do.

PR #198, opened 2017-09-05
https://github.com/hmrc/gform/pull/141

lutnos:


test/uk/gov/hmrc/gform/services/NumberValidationSpec.scala
```diff
@@ -46,6 +46,22 @@ class NumberValidationSpec extends FlatSpec with Matchers with EitherMatchers wi
     result.toEither should beRight(())
   }
 
+  "Number format" should "accept stirling pound and commas within numbers" in {
+    val textConstraint = Number()
+    val number = Text(textConstraint, Constant(""))
+
+    val fieldValue = FieldValue(FieldId("n"), number,
```
pawelpanasewicz: There is an `ExampleData` trait which is mixed into `Spec` - the common spec trait for our unit tests. It might be beneficial to reuse existing code form `ExampleData` which we tend to do. 
It's possible to either override particular fields from `ExampleData` or create another one there (I encourage other option since we don't have example `FieldValue` representing money field). 

By having such approach it's easier later to mantain code - when model changes it's easier to change only few fields in `ExampleData` and possible than iterating over all test and change their local values.

pawelpanasewicz:
Well done, thanks :+1:

PR #202, opened 2017-09-07
[EEITT-1264] - Repeating groups duplicate line breaks

paul-request:


app/uk/gov/hmrc/gform/views/group.scala.html
```diff
@@ -29,7 +28,9 @@ <h2 class="heading-small">
         </div>
       </div>
     </div>
-  } else {
+  }
```
tenochdlc: The problem with deleting that "else" is that non-repeating groups don't get a label and unless I am mistaken, they need a label too

DanielConnelly: would fieldValue.label return a boolean aswell ?

paul-request: @tenochdlc Both will get labels now, with this revised logic. But if it is not repeating and the label is empty, then it will not display it

paul-request:
@tenochdlc @DanielConnelly Thanks guys, I have amended the code. Please check again.

PR #203, opened 2017-09-07
EEITT-1238 £ sign on money text component - front end

DanielAjibike:


app/uk/gov/hmrc/gform/views/field_template_text.scala.html
```diff
@@ -25,13 +33,26 @@
...
+        @if(!isSterling) {
         <input
                 type="text"
                 name="@fieldValue.id"
                 @if(!fieldValue.editable) { readonly="" }
                 id="@fieldValue.id"
                 value="@validationResult.flatMap(_.getCurrentValue).getOrElse(prepop)"
                 class="input--medium">
+        }
+        @if(isSterling) {
```
tenochdlc: why not an "else" instead of this if?

PR #212, opened 2017-09-08
Pawel/validations wip

pawelpanasewicz:


pawelpanasewicz:
i've rebased and published above changes as new RP

PR #246, opened 2017-09-18
capturing correct customerId and passing it back to backend for submi…

DanielConnelly:
…ssion

### DIFF
--- a/app/uk/gov/hmrc/gform/auth/AuthService.scala
+++ b/app/uk/gov/hmrc/gform/auth/AuthService.scala
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2017 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


app/uk/gov/hmrc/gform/auth/AuthService.scala
```diff
@@ -0,0 +1,46 @@
...
+package uk.gov.hmrc.gform.auth
+
+import uk.gov.hmrc.auth.core.authorise.AffinityGroup
+import uk.gov.hmrc.gform.auth.models.Retrievals
+import uk.gov.hmrc.gform.prepop.AuthContextPrepop
+import uk.gov.hmrc.gform.sharedmodel.formtemplate._
+
+class AuthService {
+
+  def eeitReferenceNumber(retrievals: Retrievals) = retrievals.userDetails.affinityGroup match {
```
pawelpanasewicz: please add return type to method signature

PR #286, opened 2017-09-28
adding in config for this future timeout to make changes to it withou…

DanielConnelly:
…t making code changes


conf/application.conf
```diff
@@ -194,3 +194,4 @@ Prod {
 # false -> uses the legacy GG end point
 enrolment-service.use-tax-enrolments=false
 whitelisting.users= "test@test.com"
+future.timeout.int=20
```
lutnos: I don't think you meant to do that?  But: future.timeout=20

lutnos: Oops, my bad

PR #289, opened 2017-09-28
User friendly pdf mark eeitt 1364

HarrisonBaxter:


app/uk/gov/hmrc/gform/auditing/AuditService.scala
```diff
@@ -112,4 +111,23 @@ trait AuditService {
     )
   }
 
+  def sendSubmissionEventHashed(hashedValue: String)(implicit hc: HeaderCarrier, ec: ExecutionContext) = {
```
lutnos: The submission event needs to include the hashedvalue string as well as the hash

PR #300, opened 2017-10-03
[EEITT-1477] Inline group remove button fix

paul-request:


paul-request:
Also needs correct format in corresponding templates: https://github.com/hmrc/gform/pull/190

PR #303, opened 2017-10-03
repeating group rendering fix still needs idea how to properly render…

DanielConnelly:
… multiline group components


pawelpanasewicz:
Looks good! Please resolve conflicts and make it part of main party :+1:

PR #304, opened 2017-10-03
changes to way group is rendered.

DanielConnelly:


DanielConnelly:
Still needs paul to have a look.

PR #337, opened 2017-10-26
Continue or delete progress pages

paul-request:


app/uk/gov/hmrc/gform/views/hardcoded/pages/continue_form_page.scala.html
```diff
@@ -19,37 +19,42 @@
 
 
 @uk.gov.hmrc.gform.views.html.main_template(
-    title = "Continue Where you left Off",
+    title = "Confirm save or delete",
     frontendAppConfig = frontendAppConfig
 ) {
-<h1>Form Started, Continue or Delete and Start again</h1>
+<h1>Would you like to continue your claim or start again?</h1>
```
lutnos: I see the content says "claim".  Don't we need to deal with clams & returns, depending on the for category?

paul-request:
@lutnos good spot, Daniel has stepped in to fix that

PR #392, opened 2018-05-04
GFC-287 Prevent hitting enter adding/removing a tax type section

thebailers:


lutnos:
Simon

But I think this prevents users from adding sections unless they have JavaScript enabled?

We have to retain the ability to complete the journey with JavaScript disabled

app/uk/gov/hmrc/gform/views/form/snippets/group.scala.html
```diff
@@ -54,7 +54,8 @@ <h2 class="heading-small heading-section">
 
   @if(groupField.repeatsMax.isDefined && !disableAddButton) {
     <button
-      type="submit"
+      type="button"
```
lutnos: Simon

But I think this prevents users from adding sections unless they have JavaScript enabled?

We have to retain the ability to complete the journey with JavaScript disabled

thebailers:
Hi Dave - I have updated the code - the button types remain as submit, and then JS is used to set the attribute to 'button'. 
Then on focus of the elements, the type is updated to submit and back to button on blur.

PR #396, opened 2018-05-09
Evaluate eeitt context variables from the eeitt service

lutnos:
Evaluate eeitt context variables from the eeitt service for customeId at declaration

### DIFF
--- a/app/uk/gov/hmrc/gform/auditing/AuditService.scala
+++ b/app/uk/gov/hmrc/gform/auditing/AuditService.scala
@@ -64,26 +64,26 @@ trait AuditService {
     processedData.map(x => x.id.value -> x.value).toMap
   }
 
-  def sendSubmissionEvent(form: Form, sections: List[BaseSection], retrievals: Retrievals)(implicit ec: ExecutionContext, hc: HeaderCarrier, request: Request[_]) = {
-    sendEvent(form, formToMap(form, sections), retrievals)
+  def sendSubmissionEvent(form: Form, sections: List[BaseSection], retrievals: Retrievals, customerId: String)(implicit ec: ExecutionContext, hc: HeaderCarrier, request: Request[_]) = {
+    sendEvent(form, formToMap(form, sections), retrievals, customerId)
   }
 
-  private def sendEvent(form: Form, detail: Map[String, String], retrievals: Retrievals)(implicit ec: ExecutionContext, hc: HeaderCarrier, request: Request[_]): String = {
-    val event = eventFor(form, detail, retrievals)
+  private def sendEvent(form: Form, detail: Map[String, String], retrievals: Retrievals, customerId: String)(implicit ec: ExecutionContext, hc: HeaderCarrier, request: Request[_]): String = {
+    val event = eventFor(form, detail, retrievals, customerId)
     auditConnector.sendExtendedEvent(event)
     event.eventId
   }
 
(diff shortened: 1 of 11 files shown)


app/uk/gov/hmrc/gform/auditing/AuditService.scala
```diff
@@ -64,26 +64,26 @@ trait AuditService {
     processedData.map(x => x.id.value -> x.value).toMap
   }
 
-  def sendSubmissionEvent(form: Form, sections: List[BaseSection], retrievals: Retrievals)(implicit ec: ExecutionContext, hc: HeaderCarrier, request: Request[_]) = {
-    sendEvent(form, formToMap(form, sections), retrievals)
+  def sendSubmissionEvent(form: Form, sections: List[BaseSection], retrievals: Retrievals, customerId: String)(implicit ec: ExecutionContext, hc: HeaderCarrier, request: Request[_]) = {
```
josef-vlach: This `customerId` is great candidate for lifting into value class, so we avoid working with `String` type. I'd be happy to see this change, but it does not necessary needs to be part of this PR.

app/uk/gov/hmrc/gform/auth/AuthService.scala
```diff
@@ -32,17 +35,19 @@ class AuthService {
       .fold("")(_.getIdentifier(EEITTAuthConfig.nonAgentIdName).fold("")(_.value))
   }
 
-  def evaluateSubmissionReference(expression: TextExpression, retrievals: Retrievals, data: Map[FormComponentId, Seq[String]]): String = {
+  def evaluateSubmissionReference(expression: TextExpression, retrievals: Retrievals, formTemplate: FormTemplate,
+    data: Map[FormComponentId, Seq[String]])(implicit hc: HeaderCarrier): Future[String] = {
```
josef-vlach: So this method returns `customerId`, but it is really hard to see. Once we would introduced `CustomerId` value class, return method `Future[CustomerId]` would be more pleasant to read and understand.

app/uk/gov/hmrc/gform/gform/AcknowledgementController.scala
```diff
@@ -87,49 +88,52 @@ class AcknowledgementController(
     submissionReference: Option[TextExpression],
     retrievals: Retrievals,
     hashedValue: String,
+    formTemplate: FormTemplate,
     data: Map[FormComponentId, Seq[String]]
-  ): String = {
+  )(implicit hc: HeaderCarrier): Future[String] = {
```
josef-vlach: At the return from this method we are unnecessary throwing away information we are returning `org.jsoup.nodes.Document`. We may hold this information little longer and convert to `String` when necessary later.

app/uk/gov/hmrc/gform/auditing/AuditService.scala
lutnos: Agreed, created https://jira.tools.tax.service.gov.uk/browse/GFC-337

app/uk/gov/hmrc/gform/gform/AcknowledgementController.scala
lutnos: Yes.  Further, we should not use jsoup like that at all.  HTML should be generated correctly from views, not generated and then edited.

PR #397, opened 2018-05-10
Eliminate unnecessary calls of keystore cache fetch

lutnos:


app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -70,7 +70,9 @@ class SectionRenderingService(
     dynamicSections: List[BaseSection],
     formMaxAttachmentSizeMB: Int,
     section: BaseSection,
-    retrievals: Retrievals
+    retrievals: Retrievals,
+    repeatCache: Future[Option[CacheMap]],
```
josef-vlach: This is not nice. Is there some reason why we cannot resolve this `Future` before we create `ExtraInfo` instance?

In general we are threading `Future[Option[CacheMap]]` everywhere causing us problem with `Await` later on. It would be much better to do
```scala
for {
  cache <- repeatService.getCache
  xxx <- doAsyncCall(cache) // def doAsyncCall(cache: Option[CacheMap])
} yield ...
```
instead
```scala
for {
  xxx <- doAsyncCall(repeatService.getCache)  // def doAsyncCall(cache: Future[Option[CacheMap]])
} yield ...
```

PR #490, opened 2018-09-04
Jack/gfc 558

WitchellJack:


app/uk/gov/hmrc/gform/views/form/snippets/field_template_text_total.scala.html
```diff
@@ -29,6 +29,18 @@
...
+    val zero = BigDecimal("0")
+    val currentAmount = BigDecimal(current)
+    if (currentAmount < zero){
+        val finalNegativeAmount = "-£" + currentAmount.abs.toString()
+        finalNegativeAmount
+    } else {
+        val finalPositiveAmount = "£" + current
+        finalPositiveAmount
+    }
+}
```
josef-vlach: This seems to be broken in two ways.

1. We are adding `£` even for non Sterling values. (I do not see equivalent logic of removed line 53)
2. In other places of application we took approach to use `NumberFormat.getCurrencyInstance(Locale.UK)` to format Sterling values. We should use it here as well. See [TextFormatter.scala](https://github.com/hmrc/gform-frontend/blob/master/app/uk/gov/hmrc/gform/views/summary/TextFormatter.scala#L33) for inspiration.

WitchellJack: @josef-vlach  I believe this should fix all reported issues?

WitchellJack:
@josef-vlach This should now be finished? I have tested this in local and it works.

josef-vlach:
Looks good. You need to rebase to resolve conflicts and please squash commits into one (no need to keep interim commits in history)

lutnos:
Fine @josef-vlach , we normally merge Jack's commits, so if he does the rebase then. we can do the squash when we merge.  

@WitchellJack yes, please rebase, you do not need to squash, we can do that when we merge it.

PR #510, opened 2018-09-17
Gfc 564 add field descriptor to default errors

nubz:


nubz:
need to update unit test messages to match!

PR #515, opened 2018-09-24
GFC-580: added changebutton to group fields

nubz:


nubz:
As requested I've changed the signature for changeButton to pass the whole fieldValue and let the twirl template do the editable logic instead.

lutnos:
I have also created a story [Summary Change button refactor](https://jira.tools.tax.service.gov.uk/browse/GFC-601)

PR #541, opened 2018-11-16
GFC-14 - added new method for GG only authentication

josef-vlach:


nubz:
Tested new method on agent enrolment prologue and works fine.

PR #545, opened 2018-11-28
Gfc 654 - file upload accessibility changes & refactor

thebailers:


app/uk/gov/hmrc/gform/views/form/snippets/field_template_file_upload.scala.html
```diff
@@ -24,70 +24,73 @@
...
-                    @defaultLabel
-                }
-            </a>
-          </li>
-        </ul>
-    </div>
-</div>
+  <div id="@fileId-files" class="@{"uploaded-files" + {if(currentValue.nonEmpty) " subsection" else "" }}">
+    @if(currentValue.nonEmpty) {
+      <span href="#">@{currentValue.get}</span>
```
josef-vlach: Calling `.get` on `Option[X]` values is bug waiting to happen. It would be better to use `map(cv => ...).getOrElse("")` approach.

thebailers: Thanks Josef - I have pushed those changes.

PR #548, opened 2018-12-03
GFC-685: added handlers for legacy forms

nubz:


app/uk/gov/hmrc/gform/gform/FormController.scala
```diff
@@ -81,11 +81,15 @@ class FormController(
...
-      (cache.formTemplate.draftRetrievalMethod, cache.retrievals.affinityGroup) match {
-        case (Some(FormAccessCodeForAgents), Some(AffinityGroup.Agent)) =>
-          Future.successful(Ok(access_code_start(cache.formTemplate, AgentAccessCode.form, lang, frontendAppConfig)))
-        case _ => Future.successful(Redirect(routes.FormController.newForm(formTemplateId, lang)))
-      }
+      val formId = FormId(cache.retrievals.userDetails, formTemplateId, None)
+      for {
+        maybeForm <- gformConnector.maybeForm(formId)
+      } yield
+        (cache.formTemplate.draftRetrievalMethod, cache.retrievals.affinityGroup, maybeForm.isDefined) match {
```
lutnos: Actually, using isDefined here is not a good pattern, sorry.  Better to match mayBeForm with Some(_)

PR #549, opened 2018-12-04
GFC-668 Accessibility fixes on choice component

thebailers:


nubz:
These fixes are incorporated in GFC-579

PR #552, opened 2018-12-05
lcw/gfc 573

lancewalton:
This has already been reviewed by Josef. We need to make sure the acceptance tests are working. See the JIRA issue for details.


lancewalton:
I've just copied the changes from gform to here as well.

I agree that the copying of source is odd. If we *have* to keep things in sync (which I think is the case between gform & gform-frontend), let's just create a new module & put the shared code there.

I don't really understand enough about where the uploaded template JSON is being parsed into the Validator types. I assumed it was happening in gform-frontend.

I understand that the separation used to be due to access to mongo, but that is not the case any more. Maybe a picture would help :-)

josef-vlach:
I've discussed it with Dave and all this automation pile of work is good excuse for moving template processing and storing from `gform` to `gform-frontend` which is by far the best solution.

Under this condition we may assume this ticket to be a beginning of that movement. So please ignore my previous comment 👍

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Validator.scala
```diff
@@ -0,0 +1,72 @@
...
+    (json \ "validatorName").as[String] match {
+      case "hmrcUTRPostcodeCheck"    => json.validate[HMRCUTRPostcodeCheckValidator]
+      case "bankAccountModulusCheck" => json.validate[BankAccoutnModulusCheck]
+      case unsupported               => JsError("Unsupported '" + unsupported + "' kind of validator.")
+    }
+  }
+
+  private val reads = (basic: Reads[Validator]) | templateReads
+
+  implicit val format: OFormat[Validator] = OFormat(reads, basic)
```
josef-vlach: We can use `OFormatWithTemplateReadFallback` here as well

lancewalton:
OK. Cool. If you're happy, then we can let the QA people try it out with real templates.

project/FrontendBuild.scala
```diff
@@ -11,12 +11,16 @@ object FrontendBuild extends Build with MicroService {
...
 
+  val parsebackVersion = "0.3"
+
   val compile = Seq(
     ws,
     "uk.gov.hmrc" %% "frontend-bootstrap" % "10.7.0",
     "uk.gov.hmrc" %% "auth-client" % "2.16.0-play-25",
     "uk.gov.hmrc" %% "domain" % "5.2.0",
     "uk.gov.hmrc" %% "play-partials" % "6.1.0",
+    "com.codecommit" %% "parseback-core" % parsebackVersion,
```
josef-vlach: I had to add 
```
.settings(resolvers ++= Seq(
   Resolver.jcenterRepo,
   "bintray-djspiewak-maven" at "https://dl.bintray.com/djspiewak/maven"
))
```
to `MicroService.scala` to make this branch compile.

lancewalton: That's weird. It worked for me. I've added the resolvers anyway.

josef-vlach: I'm using coursier, so that may be the reason why it failed for me 🤔

PR #564, opened 2019-01-04
David+daryl/gfc 710/crn text format

Davidrymer97:


app/uk/gov/hmrc/gform/validation/ValidationService.scala
```diff
@@ -355,6 +356,17 @@ class ComponentsValidator(
     }
   }
 
+  private def checkCrn(fieldValue: FormComponent, value: String) = {
+    val ValidAlphaNumeric = "[A-Z]{2}[0-9]{6}".r
```
lutnos: These can be done by a single regular expression: "([A-Z]{2})?[0-9]{6}".r

PR #566, opened 2019-01-07
Daryl david/582/uk postcode formating 3

DarylMConway:


app/uk/gov/hmrc/gform/validation/ValidationService.scala
```diff
@@ -503,7 +503,7 @@ class ComponentsValidator(
       case value :: Nil =>
         Map(fieldId -> errors(fieldValue, "must not be entered")).invalid
       case value :: rest =>
-        Map(fieldId -> errors(fieldValue, "must not be entered")).invalid // we don't support multiple values yet
+        Map(fieldId -> errors(fieldValue, "must not be entered")).invalid // we don't support multiple values yetcompile
```
josef-vlach: `yetcompile` seems like typo.

app/uk/gov/hmrc/gform/validation/ValidationService.scala
```diff
@@ -522,6 +522,13 @@ class ComponentsValidator(
     }
   }
 
+  private def postcodeValidation(fieldValue: FormComponent, fieldId: FormComponentId)(xs: Seq[String]): ValidatedType =
+    (xs.filterNot(_.isEmpty), fieldId) match {
```
josef-vlach: No need to pattern match on `fieldId` as it is not used.

DarylMConway: i have removed unused feildId within the pattern matching

app/uk/gov/hmrc/gform/validation/ValidationService.scala
DarylMConway: typo has been removed

PR #572, opened 2019-01-09
GFC-719 associate help text with form control

thebailers:


nubz:
We need to change a few of these template updates - non-grouped fields should have the aria label in the input itself (text, textarea, file) - only grouped fields (e.g. choice, address, sort-code, date) should put the aria-describedby into the fieldset

PR #573, opened 2019-01-09
Daryl/720/test-only-proxy-save4later

DarylMConway:


app/uk/gov/hmrc/gform/testonly/TestOnlyController.scala
```diff
@@ -35,6 +35,8 @@ class TestOnlyController(
 
   def proxyToFileUpload(path: String): Action[Source[ByteString, _]] = proxy(fileUploadBaseUrl)(path)
 
+  def proxyToSave4later(path: String): Action[Source[ByteString, _]] = proxy(save4Later)(path)
```
lutnos: Please mend the camel case, making that proxyToSave4Later

PR #576, opened 2019-01-14
david/698/send email parameters 

Davidrymer97:


test/uk/gov/hmrc/gform/graph/FormTemplateBuilder.scala
```diff
@@ -19,6 +19,8 @@ package uk.gov.hmrc.gform.graph
 import uk.gov.hmrc.gform.sharedmodel.formtemplate._
 import uk.gov.hmrc.gform.sharedmodel.formtemplate.destinations.Destinations.DmsSubmission
 
+import scala.collection.immutable.List
```
josef-vlach: `scala.collection.immutable.List` is included in `scala.Predef` which is available in every scala source file by default, so there is no need to explicitly import it.

https://www.scala-lang.org/api/current/scala/Predef$.html

PR #587, opened 2019-01-22
tebbybearGFC-739 WIP: config added and exposed - still need templates to propa…

nubz:
…gate authConfig correctly


nubz:
opened in error

PR #591, opened 2019-01-26
David/574/implement date formats using first day and last day 

Davidrymer97:


app/uk/gov/hmrc/gform/models/DateExpr.scala
```diff
@@ -31,6 +31,7 @@ object DateExpr {
     case ExactDateValue(year, month, day) => DateExpr(year, month, day)
     case n @ NextDateValue(_, _)          => nextDateExpr(n)
     case p @ PreviousDateValue(_, _)      => previousDateExpr(p)
+
```
josef-vlach: We need to fix this compiler warning:
```
[warn]  [E1] app/uk/gov/hmrc/gform/models/DateExpr.scala:29:48:
[warn]       match may not be exhaustive.
[warn]       It would fail on the following inputs: FirstDayValue(_, _), LastDayValue(_, _)
[warn]         def fromDateValue(dv: DateValue): DateExpr = dv match {
```

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/DateValue.scala
```diff
@@ -25,6 +27,24 @@ final case class ExactDateValue(year: Int, month: Int, day: Int) extends DateVal
 final case class NextDateValue(month: Int, day: Int) extends DateValue
 final case class PreviousDateValue(month: Int, day: Int) extends DateValue
 
+final case class FirstDayValue(year: String, month: String) extends DateValue {
+  val day = 1
+
+}
+final case class LastDayValue(year: String, month: String) extends DateValue {
+
+  private val isLeapYear = (year: Int) =>
```
josef-vlach: `private val isLeapYear` function is defined two times in the code base. We need to have only one implementation to keep our code base maintainable.

lutnos: ... and doesn't you having to work it out for yourself when using an elaborate temporal library not ring alarm bells.  I believe you can use java.time.Year#isLeap(long)

lutnos:
@PasGat continues work on this

PR #594, opened 2019-01-28
Change emailParameters to Optional in form template

Davidrymer97:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Expr.scala
```diff
@@ -37,6 +37,7 @@ final case object Value extends Expr
 object FormCtx {
   lazy val readsForTemplateJson: Reads[FormCtx] = Reads {
     case JsString(exprAsStr) =>
+      println("expr as string" + exprAsStr)
```
josef-vlach: Remove this debug print

PR #595, opened 2019-01-29
Kyle/gfc 740/form.submission reference

kylemchownhmrc:


josef-vlach:
@kylemchownhmrc can you rebase this on top of master to resolve conflict?

app/uk/gov/hmrc/gform/sharedmodel/SubmissionReferenceUtil.scala
```diff
@@ -0,0 +1,59 @@
...
+    val addHyphens = d.take(4) + "-" + d.substring(4, 8) + "-" + d.takeRight(4)
+    val e = check(addHyphens)
+    addHyphens
+  }
+
+  def check(reference: String) = {
+    val a = reference.replace("-", "")
+    val b = a.toCharArray.map(i => Integer.parseInt(i.toString, 36))
+    if ((((b(0) + b(2) + b(4) + b(6) + b(8) + b(10)) * 3) + (b(1) + b(3) + b(5) + b(7) + b(9))) % 36 == b(11)) {
+      true
```
lutnos: The if(x) then { true } else {false} is redundant

app/uk/gov/hmrc/gform/sharedmodel/SubmissionReferenceUtil.scala
```diff
@@ -0,0 +1,59 @@
...
+      : Long = toUnsigned(digest(0)) + 256 * (toUnsigned(digest(1)) + 256 * (toUnsigned(digest(2)) + 256 * (toUnsigned(
+      digest(3)) + 256 * (toUnsigned(digest(4)) + 256 * (toUnsigned(digest(5)) + 256 * (toUnsigned(digest(6)) + 256 * (toUnsigned(
+      digest(7)) % 128)))))))
+
+    val b = findDigits(x, 10, Array())
+    val fd11 = (((b(0) + b(2) + b(4) + b(6) + b(8) + b(10)) * 3) + (b(1) + b(3) + b(5) + b(7) + b(9))) % 36
+    val c = b :+ fd11
+    val d = c.map(i => "0123456789ABCDEFGHIGJLMNOPQRSTUVWXYZ".toCharArray()(i.toInt)).mkString
+    val addHyphens = d.take(4) + "-" + d.substring(4, 8) + "-" + d.takeRight(4)
+    val e = check(addHyphens)
```
lutnos: It is a good idea to write the check() function now, but you do not need to call it here.  Perhaps call it from a test?

app/uk/gov/hmrc/gform/sharedmodel/SubmissionReferenceUtil.scala
```diff
@@ -0,0 +1,59 @@
...
+    addHyphens
+  }
+
+  def check(reference: String) = {
+    val a = reference.replace("-", "")
+    val b = a.toCharArray.map(i => Integer.parseInt(i.toString, 36))
+    if ((((b(0) + b(2) + b(4) + b(6) + b(8) + b(10)) * 3) + (b(1) + b(3) + b(5) + b(7) + b(9))) % 36 == b(11)) {
```
lutnos: (b(0) + b(2) + b(4) + b(6) + b(8) + b(10)) * 3) + (b(1) + b(3) + b(5) + b(7) + b(9))) % 36 is duplication, you ought to define it just once

app/uk/gov/hmrc/gform/sharedmodel/SubmissionReferenceUtil.scala
```diff
@@ -0,0 +1,59 @@
...
+    } else {
+      findDigits(source, no - 1, ((source / pow(36L, no)) % 36).toLong +: digits)
+    }
+
+  def getSubmissionReference(envelopeId: EnvelopeId) = {
+    val digest = MessageDigest.getInstance("SHA-256").digest(envelopeId.value.getBytes()).take(8)
+    def toUnsigned(x: Byte): Long = if (x < 0) 127 - x else x
+    val x
+      : Long = toUnsigned(digest(0)) + 256 * (toUnsigned(digest(1)) + 256 * (toUnsigned(digest(2)) + 256 * (toUnsigned(
+      digest(3)) + 256 * (toUnsigned(digest(4)) + 256 * (toUnsigned(digest(5)) + 256 * (toUnsigned(digest(6)) + 256 * (toUnsigned(
```
lutnos: You will find this easier using << to shift left multiples of 8 and use &  0xff or and 0x7f to make out the unsigned part.  The combine that with say List.range(0, 6) and map over the digest byte array.  There might just be a little wrinkle over & being defined on ints but not on bytes.

app/uk/gov/hmrc/gform/sharedmodel/SubmissionReferenceUtil.scala
```diff
@@ -0,0 +1,59 @@
...
+    if (no == 0) {
+      (source % 36) +: digits
+    } else if (no == 1) {
+      findDigits(source, no - 1, ((source / 36) % 36) +: digits)
+    } else {
+      findDigits(source, no - 1, ((source / pow(36L, no)) % 36).toLong +: digits)
+    }
+
+  def getSubmissionReference(envelopeId: EnvelopeId) = {
+    val digest = MessageDigest.getInstance("SHA-256").digest(envelopeId.value.getBytes()).take(8)
```
lutnos: Unusually, it would be worthwhile to include an explanatory comment here, which states that 36^11 > 2^63 .  If it were not, then we would not be getting full significance out of the digest into our 11 digits

lutnos: Actually, I think you can use BigInt to form an integer from the array of bytes, avoiding all of this

app/uk/gov/hmrc/gform/sharedmodel/SubmissionReferenceUtil.scala
```diff
@@ -0,0 +1,59 @@
...
+    def toUnsigned(x: Byte): Long = if (x < 0) 127 - x else x
+    val x
+      : Long = toUnsigned(digest(0)) + 256 * (toUnsigned(digest(1)) + 256 * (toUnsigned(digest(2)) + 256 * (toUnsigned(
+      digest(3)) + 256 * (toUnsigned(digest(4)) + 256 * (toUnsigned(digest(5)) + 256 * (toUnsigned(digest(6)) + 256 * (toUnsigned(
+      digest(7)) % 128)))))))
+
+    val b = findDigits(x, 10, Array())
+    val fd11 = (((b(0) + b(2) + b(4) + b(6) + b(8) + b(10)) * 3) + (b(1) + b(3) + b(5) + b(7) + b(9))) % 36
+    val c = b :+ fd11
+    val d = c.map(i => "0123456789ABCDEFGHIGJLMNOPQRSTUVWXYZ".toCharArray()(i.toInt)).mkString
```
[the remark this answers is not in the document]
lutnos: Actually, you can a parseInt and toString to do most of this for you.  They can do most of the radix-36 conversion for you.  I have't looked if they are available for Long or BigInt, if not you may need to decompose into two Integers

app/uk/gov/hmrc/gform/sharedmodel/SubmissionReferenceUtil.scala
lutnos: You should be able to do this with List.range() and mapping.  Also note that you can add all of the digits together the add twice the even position digits

app/uk/gov/hmrc/gform/sharedmodel/SubmissionReferenceUtil.scala
lutnos: E.g.:

  import scala.math.BigInt
  val byteArray = BigInt(1337).toByteArray
  val int = BigInt(byteArray)

app/uk/gov/hmrc/gform/sharedmodel/SubmissionReferenceUtil.scala
lutnos: Actually, I think you can use BigInt to form an integer from the array of bytes, avoiding all of this.

E.g.:

import scala.math.BigInt
val byteArray = BigInt(1337).toByteArray
val int = BigInt(byteArray)

app/uk/gov/hmrc/gform/sharedmodel/SubmissionReferenceUtil.scala
```diff
@@ -0,0 +1,61 @@
...
+      val digitArrayWithoutCheck2 = findDigits(abc.longValue(), 10, Array())
+      val digitArray = digitArrayWithoutCheck2 :+ (calcCheckChar(digitArrayWithoutCheck2) % 36)
+      val unformattedString = digitArray.map(i => Integer.toString(i.toInt, 36)).mkString.toUpperCase
+      unformattedString.take(4) + "-" + unformattedString.substring(4, 8) + "-" + unformattedString
+        .takeRight(4)
+    } else { "" }
+  }
+
+  def calcCheckChar(digits: Array[Long]): Long =
+    ((digits(0) + digits(2) + digits(4) + digits(6) + digits(8) + digits(10)) * 3) + (digits(1) + digits(3) + digits(5) + digits(
```
lutnos: You should be able to do this with foreach & Range's

PR #603, opened 2019-02-04
GFC-753: buttons are links, context in remove link, no more blank h2s

nubz:


nubz:
unit tests need updating

PR #611, opened 2019-02-07
removed whitelist, PLEASE DO NOT MERGE YET 

DarylMConway:
PLEASE DO NOT MERGE YET

### DIFF
--- a/app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
+++ b/app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
@@ -114,7 +114,7 @@ class AuthenticatedRequestActions(
                          lang,
                          request.uri,
                          getAffinityGroup,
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
```diff
@@ -114,7 +114,7 @@ class AuthenticatedRequestActions(
                          lang,
                          request.uri,
                          getAffinityGroup,
-                         ggAuthorised(request)(authUserWhitelist(_)))
+                         ggAuthorised(request)(AuthSuccessful(_).pure[Future]))
```
lutnos: See how all calls ggAuthorised now call with the parameter authGivenRetrievals as AuthSuccessful(_).pure[Future], so that parameter is no longer required.  I suspect m most of that function collapses and can be removed

PR #613, opened 2019-02-08
Kyle/gfc 706/hmrc tax period don't get obligations at beginning and end

kylemchownhmrc:


app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
```diff
@@ -165,6 +163,31 @@ class AuthenticatedRequestActions(
       } yield result
     }
 
+  def asyncWithoutObligations(
```
josef-vlach: `asyncWithoutObligations` is almost the same like `async`. There must be way how to generalise these two into common function @kylemchownhmrc .

PR #620, opened 2019-02-12
Rebased branch for textArea character limit PLEASE DO NOT MERGE YET

WitchellJack:


app/uk/gov/hmrc/gform/views/form/snippets/field_template_textarea.scala.html
```diff
@@ -64,26 +71,22 @@
   }
 
   @inline_error_map(fieldValue, validationResult)
-
-    @if(!fieldValue.isSterling) {
-      <textarea
+    @if(maxInputLength.isDefined){
```
josef-vlach: I do not like this `if` for introducing opening `<div>`, since it is totally disconnected with `if` for closing `</div>`. It can easily get broken.

PR #630, opened 2019-02-15
GCF-755

LGoldfinch2:


LGoldfinch2:
Mirroring changes made in the gform-backend.

PR #636, opened 2019-02-19
GFC-776 remove form-group wrapper from backlink anchors

thebailers:


app/uk/gov/hmrc/gform/views/form/form.scala.html
```diff
@@ -51,9 +51,7 @@
     <span id ="progress-indicator" class="form-hint">@localisation(page.progressIndicator.get)</span>
     }
     @if(shouldDisplayBack){
-     <div class="form-group js-visible">
-      <a href="#" class="link-back" id="backButton">Back</a>
-    </div>
+        <a href="#" class="link-back" id="backButton">Back</a>
```
nubz: As we have removed the wrapping div with the js-visible class then we should apply that to the standalone link - these links do not work without JavaScript

thebailers: Thanks Ali, I have updated this now.

PR #643, opened 2019-02-25
Kyle/gfc 791/cache tax period

kylemchownhmrc:


app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,114 @@
...
-      id ->
-        TaxPeriods(obligation.map(l =>
-          TaxPeriod(l.inboundCorrespondenceFromDate, l.inboundCorrespondenceToDate, l.periodKey))))
+  def updateObligations(formId: FormId, userData: UserData, form: Form, newForm: Form)(
+    implicit hc: HeaderCarrier,
+    ec: ExecutionContext) =
+    if (form.obligations != newForm.obligations) {
+      gformConnector.updateUserData(formId, userData)
+    } else {
+      Future.successful()
```
josef-vlach: What are we returning here ???

app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
```diff
@@ -195,9 +197,17 @@ class AuthenticatedRequestActions(
...
-      result      <- f(AuthCacheWithForm(retrievals, form, formTemplate, obligations))
+      form    <- gformConnector.getForm(FormId(retrievals, formTemplate._id, maybeAccessCode))
+      newForm <- obligationService.lookupIfPossible(form, formTemplate, authService, retrievals)
+      update <- obligationService.updateObligations(
+                 form._id,
+                 UserData(newForm.formData, newForm.status, newForm.visitsIndex, newForm.obligations),
+                 form,
+                 newForm)
+      result <- {
+        update
```
josef-vlach: `update` ?

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,114 @@
...
+      listOfIdNumbers <- Future.sequence(
+                          hmrcTaxPeriodIdentifiers.map(
+                            i =>
+                              authService.evaluateSubmissionReference(
+                                i.idNumber,
+                                retrievals,
+                                formTemplate,
+                                FormDataHelpers.formDataMap(form.formData),
+                                form.envelopeId)))
+      replaceIdNumber <- Future(listOfIdNumbers.zip(hmrcTaxPeriodIdentifiers))
```
josef-vlach: No need for `Future` here. Use `=` instead of `<-`

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,114 @@
...
+    formTemplate: FormTemplate,
+    authService: AuthService,
+    retrievals: MaterialisedRetrievals,
+    form: Form)(implicit hc: HeaderCarrier, ec: ExecutionContext) = {
+    val hmrcTaxPeriodIdentifiers = formTemplate.expandFormTemplateFull.allFCs.collect {
+      case IsHmrcTaxPeriod(el) => el
+    }
+    for {
+      listOfIdNumbers <- Future.sequence(
+                          hmrcTaxPeriodIdentifiers.map(
```
josef-vlach: `Future.sequence` followed by `map` is `Future.traverse`

```
Future.traverse(hmrcTaxPeriodIdentifiers)(
                          i =>
                            authService.evaluateSubmissionReference(
                              i.idNumber,
                              retrievals,
                              formTemplate,
                              FormDataHelpers.formDataMap(form.formData),
                              form.envelopeId))
```

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,114 @@
...
+                              authService.evaluateSubmissionReference(
+                                i.idNumber,
+                                retrievals,
+                                formTemplate,
+                                FormDataHelpers.formDataMap(form.formData),
+                                form.envelopeId)))
+      replaceIdNumber <- Future(listOfIdNumbers.zip(hmrcTaxPeriodIdentifiers))
+      output <- gformConnector
+                 .getAllTaxPeriods(replaceIdNumber.map(a =>
+                   HmrcTaxPeriod(a._2.idType, TextExpression(Constant(a._1)), a._2.regimeType)))
```
josef-vlach: Can we name what is `a._1` and `a._2`?

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,114 @@
...
+                                     i =>
+                                       authService.evaluateSubmissionReference(
+                                         i.idNumber,
+                                         retrievals,
+                                         formTemplate,
+                                         FormDataHelpers.formDataMap(form.formData),
+                                         form.envelopeId)))
+                     output <- {
+                       if (idNumbers.forall(i => string.equals(i))) {
+                         Future(form)
```
josef-vlach: Use `Future.successful`

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,114 @@
...
+                 hmrcTaxPeriodIdentifiers.head.idNumber,
+                 retrievals,
+                 formTemplate,
+                 FormDataHelpers.formDataMap(form.formData),
+                 form.envelopeId)
+      output <- form.obligations match {
+                 case Some(x) => {
+                   for {
+                     idNumbers <- Future.sequence(
+                                   x.listAllInfo.map(
```
josef-vlach: Again `sequence ... map` is `traverse`

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
```diff
@@ -61,3 +62,41 @@ case class TaxResponse(id: HmrcTaxPeriod, obligation: Obligation)
...
+object AllInfo {
+  implicit val format: OFormat[AllInfo] = derived.oformat
+}
+
+case class ListAllInfo(listAllInfo: List[AllInfo])
+
+object ListAllInfo {
+  implicit val format: OFormat[ListAllInfo] = Json.format[ListAllInfo]
+
+  implicit val optionFormat: OFormat[Option[ListAllInfo]] = new OFormat[Option[ListAllInfo]] {
```
josef-vlach: Why do we need custom `OFormat`?

app/uk/gov/hmrc/gform/sharedmodel/form/UserData.scala
```diff
@@ -16,16 +16,33 @@
...
+import play.api.libs.json.{ Json, OFormat, OWrites, Reads }
+import uk.gov.hmrc.gform.sharedmodel.{ ListAllInfo }
+import play.api.libs.functional.syntax._
 
 case class UserData(
   formData: FormData,
   formStatus: FormStatus,
-  visitsIndex: VisitIndex
+  visitsIndex: VisitIndex,
+  obligations: Option[ListAllInfo]
```
josef-vlach: Why do we need an `Option`?

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
kylemchownhmrc: The default OFormat doesn't work with Options as far as I can tell

app/uk/gov/hmrc/gform/sharedmodel/form/UserData.scala
kylemchownhmrc: Sometimes we don't have the information needed to make the API call at the start of the form. As such obligations would be None at that point

app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
kylemchownhmrc: The reason update was included here was to ensure that updateObligations actually finishes before moving on

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
kylemchownhmrc: The idea was that as updateUserData returns a Future[Unit] the else needed to return the same

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
kylemchownhmrc: I'm not to sure what you mean. They were zipped just above this and naming them by making them different variables would defeat the point of doing that

app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
josef-vlach: That is guaranteed by monad laws ie. by flatMap method.

app/uk/gov/hmrc/gform/auth/AuthService.scala
```diff
@@ -224,6 +224,8 @@ class AuthService(
 
       case SubmissionReference => getSubmissionReference(envelopeId).pure[Future]
 
+      case Constant(value) => Future.successful(value)
```
josef-vlach: Seeing this now in bigger context, we need to completely get rid of this whole method `evaluateSubmissionReference`.
It is duplicating (incompletely) what `eval` on `uk.gov.hmrc.gform.graph.Evaluator` is for. Is there something preventing us to use this `eval` instead?

kylemchownhmrc: Not that I know of. I wasn't aware that Evaluator even existed. Is removing evaluateSubmissionReference something for me to do or to leave for another ticket ?

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,105 @@
...
       case IsHmrcTaxPeriod(el) => el
     }
-    val futureListOfTaxPeriodDes =
-      gformConnector
-        .getAllTaxPeriods(hmrcTaxPeriodIdentifiers)
-        .map(i => i.flatMap(j => j.obligation.obligations.flatMap(h => makeMap(j.id, h.obligationDetails))))
-    futureListOfTaxPeriodDes.map(i => i.toMap)
+    for {
+      string <- authService.evaluateSubmissionReference(
+                 hmrcTaxPeriodIdentifiers.head.idNumber,
```
josef-vlach: `head` on `List` is unsafe, we do not want to use it.

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,105 @@
...
     val hmrcTaxPeriodIdentifiers = formTemplate.expandFormTemplateFull.allFCs.collect {
       case IsHmrcTaxPeriod(el) => el
     }
-    val futureListOfTaxPeriodDes =
-      gformConnector
-        .getAllTaxPeriods(hmrcTaxPeriodIdentifiers)
-        .map(i => i.flatMap(j => j.obligation.obligations.flatMap(h => makeMap(j.id, h.obligationDetails))))
-    futureListOfTaxPeriodDes.map(i => i.toMap)
+    for {
+      string <- authService.evaluateSubmissionReference(
```
josef-vlach: `string` is not too descriptive.

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
josef-vlach: The thing is you are not passing `()` as the parameter to the `successful` method, do you see it?

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
josef-vlach: In couple of days as code grows, refactoring happen etc.
> They were zipped just above this ...
may become completely different part of code base so reader will have no clue whatsoever what `replaceIdNumber` represent.

It would be much nicer if mapping over `replaceIdNumber` happen straight after zipping.

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,105 @@
...
+                     idNumbers <- Future.traverse(x.listAllInfo)(
+                                   i =>
+                                     authService.evaluateSubmissionReference(
+                                       i.hmrcTaxPeriod.idNumber,
+                                       retrievals,
+                                       formTemplate,
+                                       FormDataHelpers.formDataMap(form.formData),
+                                       form.envelopeId))
+                     output <- {
+                       if (idNumbers.forall(i => string.equals(i))) {
```
josef-vlach: Let's use `===` from cats instead of `equals`

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,105 @@
...
+                         Future.successful(form)
+                       } else {
+                         val newObligations = lookupObligationsMultiple(formTemplate, authService, retrievals, form)
+                         newObligations.map(i => form.copy(obligations = Some(ListOfTaxPeriodInformation(i))))
+                       }
+                     }
+                   } yield output
+                 }
+                 case None => {
+                   if (hmrcTaxPeriodIdentifiers.forall(i => !i.regimeType.value.isEmpty && !string.isEmpty)) {
```
josef-vlach: This `if ... else` is that same (except condition) as the `if ... else` starting on line 92. Let's reuse the one code for both cases.

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
josef-vlach: > doesn't work
What you mean by that? It doesn't compile, it works incorrectly or something else? We use `Option` all over the place...

app/uk/gov/hmrc/gform/sharedmodel/form/UserData.scala
josef-vlach: But what is difference between `None` and empty `List` in that case?

app/uk/gov/hmrc/gform/auth/AuthService.scala
josef-vlach: No need for another ticket, it is work which can safely be done as part of this ticket.

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,108 @@
...
       case IsHmrcTaxPeriod(el) => el
     }
-    val futureListOfTaxPeriodDes =
-      gformConnector
-        .getAllTaxPeriods(hmrcTaxPeriodIdentifiers)
-        .map(i => i.flatMap(j => j.obligation.obligations.flatMap(h => makeMap(j.id, h.obligationDetails))))
-    futureListOfTaxPeriodDes.map(i => i.toMap)
+    for {
+      currentIdNumber <- authService.evaluateSubmissionReference(
+                 hmrcTaxPeriodIdentifiers match {
```
josef-vlach: This pattern match is ok, but it should be done as first thing (ie. on line 73) since if the list is empty we know there is no work to do and we can immediately return.

app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
```diff
@@ -195,9 +197,17 @@ class AuthenticatedRequestActions(
     maybeAccessCode: Option[AccessCode],
     formTemplate: FormTemplate)(retrievals: MaterialisedRetrievals)(implicit hc: HeaderCarrier): Future[Result] =
     for {
-      form        <- gformConnector.getForm(FormId(retrievals, formTemplate._id, maybeAccessCode))
-      obligations <- obligationService.lookupObligationsMultiple(formTemplate)
-      result      <- f(AuthCacheWithForm(retrievals, form, formTemplate, obligations))
+      form    <- gformConnector.getForm(FormId(retrievals, formTemplate._id, maybeAccessCode))
+      newForm <- obligationService.lookupIfPossible(form, formTemplate, authService, retrievals)
+      update <- obligationService.updateObligations(
```
josef-vlach: Why are we calling `updateObligations` here? It seems like completely wrong thing to do. Why aren't we doing this logic in `FormController`?

lutnos: @kylemchownhmrc , updateObligations is the last thing called before f and calling it in AuthenticatedRequests is not ready early appropriate.  That function f has been passed in (via async) from the controller.

So it would be more appropriate to move the call out to the beginning of the functions for, e.g. uk.gov.hmrc.gform.gform.FormController#form and uk.gov.hmrc.gform.gform.SummaryController#submit .  

Note that in some of the uses, e.g. DeclarationController and AcknowledgementController it is actually inappropriate to update the userData, because the form has already been finally validated or indeed submitted.  They can actually call uk.gov.hmrc.gform.controllers.AuthenticatedRequestActions#asyncWithoutObligations instead.  You can almost get away with asyncWithoutObligations in most or all cases except for the two above, so leaving just two calls to updateObligations, one for form sections and the second for summary, where final validation occurs.

You will just need to add form to uk.gov.hmrc.gform.controllers.AuthCacheWithForm , so that the functions called can call updateObligations with form as well as newForm.

Do you see?

app/uk/gov/hmrc/gform/auth/AuthService.scala
kylemchownhmrc: How would you deal with the fact that `eval` requires imports that `evaluateSubmissionReference` doesn't? There are things that `eval` needs that we don't have access to where `evaluateSubmissionReference` is called

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,131 @@
...
+                          i =>
+                            authService.evaluateSubmissionReference(
+                              i.idNumber,
+                              retrievals,
+                              formTemplate,
+                              FormDataHelpers.formDataMap(form.formData),
+                              form.envelopeId))
+      output <- gformConnector
+                 .getAllTaxPeriods(
+                   NonEmptyList.fromList(
```
josef-vlach: This is unsafe there is no guarantee that list we are passing into `fromList` is not `Nil`

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,131 @@
...
+                         idNumbers <- Future.traverse(x.toList)(
+                                       i =>
+                                         authService.evaluateSubmissionReference(
+                                           i.hmrcTaxPeriod.idNumber,
+                                           retrievals,
+                                           formTemplate,
+                                           FormDataHelpers.formDataMap(form.formData),
+                                           form.envelopeId))
+                         output <- {
+                           ifStatement(
```
josef-vlach: This `ifStatement` would be much better when defined as local method in `lookupIfPossible`. Than all it need to take is `condition: Boolean` as a parameter. All the rest is available in scope.

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
```diff
@@ -61,3 +65,29 @@ case class TaxResponse(id: HmrcTaxPeriod, obligation: Obligation)
...
+  inboundCorrespondenceToDate: Date,
+  periodKey: String)
+
+object TaxPeriodInformation {
+  implicit val format: OFormat[TaxPeriodInformation] = derived.oformat
+}
+
+sealed trait Obligations
+final case object NotChecked extends Obligations
+final case object NoObligations extends Obligations
```
josef-vlach: Where we create this `NoObligations` case? It seems it is not needed after all.

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,131 @@
...
-import scala.concurrent.ExecutionContext
+
+import scala.concurrent.{ ExecutionContext, Future }
 
 class ObligationService(gformConnector: GformConnector) {
 
   val stringToDate = new SimpleDateFormat("yyyy-MM-dd")
 
-  def lookupObligationsMultiple(formTemplate: FormTemplate)(implicit hc: HeaderCarrier, ec: ExecutionContext) = {
+  def lookupObligationsMultiple(
```
josef-vlach: lets mark this method `private`

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,131 @@
...
+    formTemplate: FormTemplate,
+    authService: AuthService,
+    retrievals: MaterialisedRetrievals,
+    form: Form)(implicit hc: HeaderCarrier, ec: ExecutionContext) = {
+    val hmrcTaxPeriodIdentifiers = formTemplate.expandFormTemplateFull.allFCs.collect {
```
josef-vlach: Since `lookupObligationsMultiple` is called only from `lookupIfPossible` we can passed in `hmrcTaxPeriodIdentifiers : NonEmptyList[HmrcTaxPeriod]` as input parameter and do not calculate it second time (plus we already now that we need non empty list).

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,131 @@
...
+                              formTemplate,
+                              FormDataHelpers.formDataMap(form.formData),
+                              form.envelopeId))
+      output <- gformConnector
+                 .getAllTaxPeriods(
+                   NonEmptyList.fromList(
+                     listOfIdNumbers
+                       .zip(hmrcTaxPeriodIdentifiers)
+                       .map(a => HmrcTaxPeriod(a._2.idType, TextExpression(Constant(a._1)), a._2.regimeType))) match {
+                     case Some(x) => x
```
josef-vlach: We are not handling `None` case here.

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,131 @@
...
+                              retrievals,
+                              formTemplate,
+                              FormDataHelpers.formDataMap(form.formData),
+                              form.envelopeId))
+      output <- gformConnector
+                 .getAllTaxPeriods(
+                   NonEmptyList.fromList(
+                     listOfIdNumbers
+                       .zip(hmrcTaxPeriodIdentifiers)
+                       .map(a => HmrcTaxPeriod(a._2.idType, TextExpression(Constant(a._1)), a._2.regimeType))) match {
```
josef-vlach: Why are we passing `TextExpression` (ie. `TextExpression(Constant(a._1))`) into backend? We should pass evaluated value instead.

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,123 @@
...
+        val newObligations = lookupObligationsMultiple(formTemplate, authService, retrievals, form, hmrcTaxPeriodIdentifiersNonEmpty)
+        newObligations.map(i => form.copy(obligations = RetrievedObligations(i)))
+      } else {
+        Future.successful(form)
+      }
+
+    hmrcTaxPeriodIdentifiers match {
+      case a :: _ =>
+        for {
+          currentIdNumber <- authService.evaluateSubmissionReference(
```
josef-vlach: (Continuation form https://github.com/hmrc/gform-frontend/pull/643/files#r260219771)

All this call should never happen, nor should `eval` be called here directly (the same holds for line 48 and 109). 

What should happen instead is that `eval` should evaluate `HmrcTaxPeriod` component's `idNumber` expression (when called as part of `ProcessDataService.getProcessData`) and it should be given back to this service as part of `FormDataRecalculated` class instance. 

If we cannot get hold of `FormDataRecalculated` instance, something is wrong in whole design and bigger refactor/rethinking will be needed.

app/uk/gov/hmrc/gform/obligation/ObligationService.scala
```diff
@@ -17,30 +17,123 @@
...
 
   val stringToDate = new SimpleDateFormat("yyyy-MM-dd")
 
-  def lookupObligationsMultiple(formTemplate: FormTemplate)(implicit hc: HeaderCarrier, ec: ExecutionContext) = {
+  private def lookupObligationsMultiple(
+    formTemplate: FormTemplate,
+    authService: AuthService,
+    retrievals: MaterialisedRetrievals,
+    form: Form,
+    hmrcTaxPeriodIdentifiers: NonEmptyList[HmrcTaxPeriod])(implicit hc: HeaderCarrier, ec: ExecutionContext) = {
```
josef-vlach: Wait, what are we doing here actually? It just occurred to me, but this looks like as contradioction of what I intended to achieve by including 

```
Given the formtemplate has more than 1 hmrcTaxPeriodSelector component 
when parsing this formtemplate 
then reject with explanation that only 1 hmrcTaxPeriodSelector is allowed
```

in ACs in the first place.

Having not following rule this we are only making our life more complicated.

`hmrcTaxPeriodIdentifiers: NonEmptyList[HmrcTaxPeriod]` should be just `hmrcTaxPeriodIdentifier: HmrcTaxPeriod`. Or do I miss something?

PR #644, opened 2019-02-25
David/786/sbt refactoring

Davidrymer97:


build.sbt
```diff
@@ -0,0 +1,85 @@
...
+addTestReportOption(IntegrationTest, "int-test-reports")
+testGrouping in IntegrationTest := oneForkedJvmPerTest2((definedTests in IntegrationTest).value)
+parallelExecution in IntegrationTest := false
+
+resolvers ++= Seq(
+  Resolver.bintrayRepo("jetbrains", "markdown"),
+  Resolver.jcenterRepo
+)
+resolvers ++= Seq(
+  Resolver.jcenterRepo,
```
arkowalski: duplicate?

build.sbt
```diff
@@ -0,0 +1,85 @@
...
+  "bintray-djspiewak-maven" at "https://dl.bintray.com/djspiewak/maven"
```
arkowalski: do we need this one? we probably do, but could be worth checking
you can test this by running these "rm -rf ~/.ivy2
rm -rf ~/.sbt" this would remove everything you have locally and attempt to download it again, assuring you have the right resolvers in place.

build.sbt
```diff
@@ -0,0 +1,85 @@
...
+evictionWarningOptions in update := EvictionWarningOptions.default.withWarnScalaVersionEviction(false)
+routesImport ++= Seq(
+  "uk.gov.hmrc.gform.binders.ValueClassBinder._",
+  "uk.gov.hmrc.gform.sharedmodel.formtemplate._",
+  "uk.gov.hmrc.gform.sharedmodel.form._",
+  "uk.gov.hmrc.gform.sharedmodel._",
+  "uk.gov.hmrc.gform.auth._",
+  "uk.gov.hmrc.gform.models._"
+)
+//TODO enable it and fix errors in code!!!
```
arkowalski: what errors? is this still applicable? perhaps we can remove this comment? see if this is still required, if is see what the errors are about and add some key words from the logs

build.sbt
```diff
@@ -0,0 +1,85 @@
...
+import play.sbt.routes.RoutesKeys.routesImport
+import sbt.Tests.{ Group, SubProcess }
+import uk.gov.hmrc.DefaultBuildSettings._
+import uk.gov.hmrc.SbtAutoBuildPlugin
+import uk.gov.hmrc.sbtdistributables.SbtDistributablesPlugin
+import uk.gov.hmrc.sbtdistributables.SbtDistributablesPlugin.publishingSettings
+import uk.gov.hmrc.versioning.SbtGitVersioning
+import uk.gov.hmrc.versioning.SbtGitVersioning.autoImport.majorVersion
+
+lazy val plugins: Seq[Plugins] = Seq.empty
```
arkowalski: plugins and playSettings seem odd here.. 

do we even need these?

build.sbt
arkowalski: also we could have the resolvers within one "resolvers ++= Seq"

build.sbt
josef-vlach: Yes, this is still applicable. If you wonder what errors, do as comments ask and see for yourself :)

josef-vlach: To enable this by default we would need first to get rid of all warnings we have now, but that is not so simple.

PR #645, opened 2019-02-25
GFC-800: Wrapped result of displaySterling function in BigNumber 

LGoldfinch2:
Stops the error on the 'Bingo duty profits in the return period' page, ensuring the commas are in the right place and the decimal point appears. Need to clarify whether the rounding is acceptable.

### DIFF
--- a/app/uk/gov/hmrc/gform/models/helpers/Javascript.scala
+++ b/app/uk/gov/hmrc/gform/models/helpers/Javascript.scala
@@ -92,7 +92,7 @@ object Javascript {
          |};
          |function displaySterling(result) {
          |
-         |  return result < 0 ? result.replace("-", "-£") : '£' BigNumber(result).toFormat(2)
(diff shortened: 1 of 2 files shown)


app/uk/gov/hmrc/gform/models/helpers/Javascript.scala
```diff
@@ -91,7 +91,8 @@ object Javascript {
          |  return BigNumber(a).times(BigNumber(b));
          |};
          |function displaySterling(result) {
-         |  return result < 0 ? result.replace("-", "-£") : '£' + result
+         |
+         |  return result < 0 ? result.replace("-", "-£") : '£' BigNumber(result).toFormat(2)
```
nubz: we are missing the + symbol concatenation, required after the '£' to make a single string - this throws a JS error for unexpected identifier

PR #657, opened 2019-02-28
Allows you to have multiple hmrcTaxPeriod components in the same form

kylemchownhmrc:


app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
```diff
@@ -20,8 +20,12 @@ import java.util.Date
 
 import cats.data.NonEmptyList
 import julienrf.json.derived
+import play.api.data.validation.ValidationError
 import play.api.libs.json._
 import uk.gov.hmrc.gform.sharedmodel.formtemplate._
+import play.api.libs.functional.syntax._
```
josef-vlach: Unused import

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
```diff
@@ -20,8 +20,12 @@ import java.util.Date
...
+
+import scala.collection.immutable.{ ::, List, Nil }
```
josef-vlach: This import is not needed.

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
```diff
@@ -84,9 +89,35 @@ object TaxPeriodInformation {
...
 
 object Obligations {
   import JsonUtils._
   implicit val format: OFormat[Obligations] = derived.oformat
 }
+
+case class IdNumberValue(value: String) extends AnyVal
+
+object IdNumberValue {
+  import JsonUtils._
```
josef-vlach: import not needed.

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
```diff
@@ -84,9 +89,35 @@ object TaxPeriodInformation {
 
 sealed trait Obligations
 final case object NotChecked extends Obligations
-final case class RetrievedObligations(listOfObligations: NonEmptyList[TaxPeriodInformation]) extends Obligations
+final case class RetrievedObligations(listOfObligations: List[TaxPeriodInformation]) extends Obligations
 
 object Obligations {
   import JsonUtils._
```
josef-vlach: import not needed.

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
```diff
@@ -84,9 +89,35 @@ object TaxPeriodInformation {
...
+object IdNumberValue {
+  import JsonUtils._
+  implicit val format: OFormat[IdNumberValue] = derived.oformat
+}
+
+case class HmrcTaxPeriodWithEvaluatedId(hmrcTaxPeriod: HmrcTaxPeriod, idNumberValue: IdNumberValue)
+
+object HmrcTaxPeriodWithEvaluatedId {
+
+  implicit def readsNonEmptyList[T: Reads] = Reads[NonEmptyList[T]] { json =>
```
josef-vlach: readsNonEmptyList not used.

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
```diff
@@ -84,9 +89,35 @@ object TaxPeriodInformation {
...
+    Json.fromJson[List[T]](json).flatMap {
+      case Nil     => JsError(ValidationError(s"Required at least one element. Got: $json"))
+      case x :: xs => JsSuccess(NonEmptyList(x, xs))
+    }
+  }
+
+  implicit def writesNonEmptyList[T: Writes] = Writes[NonEmptyList[T]] { v =>
```
josef-vlach: writesNonEmptyList not used.

app/uk/gov/hmrc/gform/gformbackend/GformConnector.scala
```diff
@@ -137,9 +137,9 @@ class GformConnector(ws: WSHttp, baseUrl: String) {
   //TODO move this file to gform and make it's origin there
 
   /****** Tax Period ******/
-  def getAllTaxPeriods(htps: NonEmptyList[HmrcTaxPeriod])(implicit hc: HeaderCarrier, ec: ExecutionContext) = {
+  def getAllTaxPeriods(htps: List[HmrcTaxPeriodWithEvaluatedId])(implicit hc: HeaderCarrier, ec: ExecutionContext) = {
```
josef-vlach: What was the need to revert back to `List`?

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
```diff
@@ -17,13 +17,10 @@
...
 
 import java.util.Date
-
-import cats.data.NonEmptyList
 import julienrf.json.derived
-import play.api.data.validation.ValidationError
 import play.api.libs.json._
 import uk.gov.hmrc.gform.sharedmodel.formtemplate._
-import scala.collection.immutable.{ ::, List, Nil }
+import scala.collection.immutable.{List}
```
josef-vlach: Please install this plugin https://plugins.jetbrains.com/plugin/8236-scalafmt to get proper source code formatting from IntelliJ or you can start to use sbt for compilation. 

When people do not use scalafmt all other devs are potentially affected.

Btw. neither `import scala.collection.immutable.{List}` is needed

kylemchownhmrc: I installed scalafmt before the most recent commit and tried to get it to format the code but it didn't seem to do anything

app/uk/gov/hmrc/gform/gformbackend/GformConnector.scala
kylemchownhmrc: Using NonEmptyLists was causing issues as in several instances we were needing to filter over the NonEmptyLists which would then give a List which we would then need to convert back to a NonEmptyList etc.

josef-vlach: Whenever one needs to filter on NonEmptyList result cannot be safely turned back into NonEmptyList. Desire so is indication that something is probably wrong with the design. 

The solution to get rid of NonEmptyList on input is least preferred one. Cannot we just keep working with List after the filter?

kylemchownhmrc: The issue is that the NonEmptyList is filtered before this point, making it a list. As such this section of the code would need to take a List rather than a NonEmptyList even if we used NonEmptyList earlier

lutnos: You should not be calling the backend via uk.gov.hmrc.gform.gformbackend.GformConnector#getAllTaxPeriods (causing an inter-microservice http call) if the list of HmrcTaxPeriodWithEvaluatedId is empty, it will have no effect and you will simply be getting back an empty list of TaxResponse at considerable cost

kylemchownhmrc: Despite not taking a NonEmptyList we still should never call the backend with an empty list as the list we pass to getAllTaxPeriods is formed from a NonEmptyList. The reason it becomes a list is that trying to traverse over the NonEmptyList to evaluate the IdNumbers doesn't seem to work with a NonEmptyList, so I have used .toList on it

lutnos: import cats.implicits._

...

hmrcTaxPeriodIdentifiers.nonEmptyTraverse(

PR #662, opened 2019-03-05
[GFC-800][AK,LG,SM] Values from sterling fields being retrieved with commas 

LGoldfinch2:


test/uk/gov/hmrc/gform/gform/FormServiceSpec.scala
```diff
@@ -0,0 +1,71 @@
...
+        ._2
+        .getCurrentValue shouldBe Some("1000.25")
+    }
+  }
+
+  it should "when FormComponent type does not equal Sterling, commas should be untouched" in {
+    forAll(genFormComponent) { formComponent =>
+      val isSterlingConstraint = IsText.unapply(formComponent).map(_.constraint.isInstanceOf[Sterling]).getOrElse(false)
+
+      isSterlingConstraint shouldBe false
```
lancewalton: This should be done as a filter on the genFormComponentSterlingConstraint rather than as an assertion

app/uk/gov/hmrc/gform/gform/FormService.scala
```diff
@@ -0,0 +1,39 @@
...
+import uk.gov.hmrc.gform.validation.{ FieldOk, FormFieldValidationResult }
+
+class FormService {
+
+  def removeCommas(formValidatedData: List[(FormComponent, FormFieldValidationResult)])
+    : List[(FormComponent, FormFieldValidationResult)] =
+    formValidatedData.map {
+      case (formComponent, formFiledValidationResult) =>
+        formComponent match {
+          case text
```
lancewalton: Can this not be done as case IsText(text)?

app/uk/gov/hmrc/gform/gform/FormController.scala
```diff
@@ -83,11 +83,16 @@ class FormController(
                    cache.form.thirdPartyData,
                    cache.formTemplate)
                    .map {
-                     case (validationResult, v, _) =>
-                       (
-                         ValidationUtil.isFormValid(validationResult.toMap),
-                         FormData(validationResult.flatMap(_._2.toFormField)),
-                         v)
+                     case (validationResult, validatedType, _) => {
```
lancewalton: Extract the body of this case into a private function. It's too big to be inline.

app/uk/gov/hmrc/gform/gform/FormService.scala
SeanMurray1: In this case we're trying to check if it is a textConstraint of type Sterling so we do the IsText in the guard.  Do you know a better way of doing this ?

app/uk/gov/hmrc/gform/gform/FormController.scala
```diff
@@ -83,14 +85,24 @@ class FormController(
...
+
                    }
     } yield formData
 
+  private def validateFormHelper(
+    validationResult: List[(FormComponent, FormFieldValidationResult)],
+    validatedType: ValidatedType[ValidationResult]): FormValidationOutcome = {
+    val isFormValid = ValidationUtil.isFormValid(validationResult.toMap)
+    val formComponents =
+      if (isFormValid) formService.removeCommas(validationResult) else validationResult
```
josef-vlach: We cannot remove commas here, since this affect calculated values and data entered by user as well. Commas needs to be removed before we generate handlebar payload.

PR #670, opened 2019-03-08
[GFC-800][AK, LG, SM] after form fields are validated remove commas f…

LGoldfinch2:
…or sterling values

GFC-800: Final tweaks to FormController, FormService + FormServiceSpec

GFC-800: Acting on pull requests.

GFC-800: Added cases for Number and Positive Number in FormService

### DIFF
--- a/app/uk/gov/hmrc/gform/gform/FormService.scala
+++ b/app/uk/gov/hmrc/gform/gform/FormService.scala
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2019 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.gform.gform
+import uk.gov.hmrc.gform.sharedmodel.formtemplate.{ FormComponent, IsText, Sterling }
+import uk.gov.hmrc.gform.validation.{ FieldOk, FormFieldValidationResult }
+import uk.gov.hmrc.gform.ops.FormComponentOps
+
+class FormService {
+
+  def removeCommas(formValidatedData: List[(FormComponent, FormFieldValidationResult)])
+    : List[(FormComponent, FormFieldValidationResult)] =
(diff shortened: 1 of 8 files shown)


test/uk/gov/hmrc/gform/gform/FormServiceSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+  lazy val genFormComponentPNConstraint: Gen[FormComponent] =
+    genFormComponent.map(e => e.copy(`type` = textPositiveNumberConstraint))
+
+  lazy val textNumberConstraint = Text(number, Expr.additionIdentityExpr)
+  lazy val textSterlingConstraint = Text(sterling, Expr.additionIdentityExpr)
+  lazy val textPositiveNumberConstraint = Text(positiveNumber, Expr.additionIdentityExpr)
+
+  lazy val sterling = new Sterling(RoundingMode.defaultRoundingMode)
+  lazy val number = new Number()
+  lazy val positiveNumber = new PositiveNumber()
```
josef-vlach: Why is everything from line 28 to 44 `lazy`

app/uk/gov/hmrc/gform/gform/FormService.scala
```diff
@@ -0,0 +1,39 @@
...
+import uk.gov.hmrc.gform.ops.FormComponentOps
+
+class FormService {
+
+  def removeCommas(formValidatedData: List[FormComponentValidation]): List[FormComponentValidation] =
+    formValidatedData.map(removeCommasHelper)
+
+  def removeCommasHelper(formValidationOutcome: FormComponentValidation) = {
+    import formValidationOutcome._
+    formComponent match {
```
josef-vlach: This pattern match would be shorter if turned into `if...else`

app/uk/gov/hmrc/gform/gform/FormService.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  def removeCommas(formValidatedData: List[FormComponentValidation]): List[FormComponentValidation] =
+    formValidatedData.map(removeCommasHelper)
+
+  def removeCommasHelper(formValidationOutcome: FormComponentValidation) = {
+    import formValidationOutcome._
+    if (formComponent.isSterling || formComponent.isPositiveNumber || formComponent.isNumber)
+      FormComponentValidation(
+        formComponent,
+        FieldOk(formComponent, formFieldValidationResult.getCurrentValue.getOrElse("").replaceAll(",", "")))
```
josef-vlach: Problem with this solution is when input `formValidationOutcome` has no `FieldOk` as its `formFieldValidationResult`. It such a case we possibly discarded error and turn it into `FieldOk` result which is dangerous.

In current implementation this problem is avoided by guarding call to `removeCommas` by check if form is valid. But for the future sake we should implement `removeCommasHelper` in such a way that it modifies `formValidationOutcome` only if it has `FieldOk` as its `formFieldValidationResult`.

PR #687, opened 2019-03-20
Gfc 751/sm/labels being swapped

SeanMurray1:


app/uk/gov/hmrc/gform/validation/ValidationService.scala
```diff
@@ -193,9 +196,12 @@ class ComponentsValidator(
...
 
-  private def validateDate(fieldValue: FormComponent, date: Date): ValidatedType[Unit] = {
+  private def validateDate(
+    fieldValue: FormComponent,
+    date: Date,
+    otherFieldValue: Option[FormComponent]): ValidatedType[Unit] = {
     val reqFieldValidResult = validateDateRequiredField(fieldValue)
-    val otherRulesValidResult = validateDateImpl(fieldValue, date)(data)
+    val otherRulesValidResult = validateDateImpl(fieldValue, date, otherFieldValue)(data)
     Monoid[ValidatedType[Unit]].combineAll(List(reqFieldValidResult, otherRulesValidResult))
```
PasGat: Maybe we could inline the two variables passed as argument as they are only used one, so you could remove brackets too saving four lines of code ;-)

app/uk/gov/hmrc/gform/validation/ValidationService.scala
```diff
@@ -250,7 +271,7 @@ class ComponentsValidator(
     concreteDate: ConcreteDate,
     offset: OffsetDate,
     data: FormDataRecalculated): Validated[GformError, Unit] =
-    validateInputDate(fieldValue, fieldValue.id, fieldValue.errorMessage, data).andThen(
+    validateInputDate(fieldValue, fieldValue.id, fieldValue.errorMessage, data, None).andThen(
```
PasGat: We seem to be calling `validateInputDate` with None at the end quite a lot - maybe and only maybe we could consider a default argument (not ideal I would normally discourage) or an overloaded function or else

app/uk/gov/hmrc/gform/validation/ValidationService.scala
SeanMurray1: changes were resolved

app/uk/gov/hmrc/gform/validation/ValidationService.scala
SeanMurray1: changes were resolved

PR #698, opened 2019-03-28
GFC-780][DC & KC] removed duplicated fields 

DarylMConway:
removed multi-field based typing within nonSubmitted as they was being submitted twice.


josef-vlach:
@DarylMConway since this is quite elusive functionality (removing it won't seemingly break anything) would it be possible to write some tests where it would be evident what this functionality achieves?

PR #700, opened 2019-03-29
GFC-?: Chipped away at ValidationService. Produced 2 new classes from it

LGoldfinch2:
GFC-?: Broke down ComponentsValidation. Still chipping

GFC-?: Extracted methods from AddressValidation. Cleaned up code.

Mainly consisted of breaking down ValidationService. Still have to write more tests and abstract more out of the larger methods.


app/uk/gov/hmrc/gform/validation/AddressValidation.scala
```diff
@@ -0,0 +1,99 @@
...
+          validateForbiddenField("postcode", fieldValue)(addressValueOf("postcode")),
+          validateRequiredField("country", localisation("Country"), fieldValue)(addressValueOf("country"))
+        )
+    }
+
+    Monoid[ValidatedType[Unit]].combineAll(validatedResult)
+  }
+}
+
+case object AddressValidation {
```
josef-vlach: `AddressValidation` should be `object` only.  

`case object` should be used only as part of `ADT` definition ([example here](https://github.com/hmrc/gform-frontend/blob/master/app/uk/gov/hmrc/gform/sharedmodel/formtemplate/AuthConfig.scala#L112-L118)), not stand alone as in this case. This stands for all standalone `case object`s in this PR.

PR #709, opened 2019-04-02
Gfc 790/sm/deal with groups

SeanMurray1:
Now deals with groups when adding information to the pdf

### DIFF
--- a/app/uk/gov/hmrc/gform/gform/AcknowledgementController.scala
+++ b/app/uk/gov/hmrc/gform/gform/AcknowledgementController.scala
@@ -120,18 +120,20 @@ class AcknowledgementController(
     val dateFormat = DateTimeFormatter.ofPattern("dd MMM yyyy")
     val formattedTime =
       s"""${submissionDetails.submittedDate.format(dateFormat)} ${submissionDetails.submittedDate.format(timeFormat)}"""
-
     val rows = List(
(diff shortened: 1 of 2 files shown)


app/uk/gov/hmrc/gform/gform/AcknowledgementController.scala
```diff
@@ -120,18 +120,20 @@ class AcknowledgementController(
...
     )
     val extraData = cya_section("Submission details", HtmlFormat.fill(rows)).toString()
-
     val declaration: List[(FormComponent, Seq[String])] = for {
-      formTemplateDecFields <- formTemplate.declarationSection.fields
-      formData              <- data.get(formTemplateDecFields.id)
-    } yield (formTemplateDecFields, formData)
+      formTemplateDecField <- formTemplate.declarationSection.fields
+      field <- formTemplateDecField.`type` match {
+                case g: Group => g.fields
```
josef-vlach: We are implementing this logic on couple of places already. We should make this into reusable public function and reuse it where possible.

Look at `Fields.flattenGroups` to start with. (There is also `def groupOrNot` in `StructuredFormDataBuilder` which should be replaced by our new function)

SeanMurray1: Used the Fields.flattenGroups

PR #712, opened 2019-04-03
Gfc project clean up

LGoldfinch2:


app/uk/gov/hmrc/gform/validation/DateValidation.scala
```diff
@@ -0,0 +1,340 @@
...
+          ComponentsValidator.validateRF(fieldValue, "month")(dateValueOf("month")),
+          ComponentsValidator.validateRF(fieldValue, "year")(dateValueOf("year"))
+        )
+      case false => List(().valid)
+    }
+    Monoid[ValidatedType[Unit]].combineAll(validatedResult)
+  }
+
+  private def validateDateImpl(fieldValue: FormComponent, date: Date, otherFieldValue: Option[FormComponent])(
+    data: FormDataRecalculated): ValidatedType[Unit] =
```
josef-vlach: This `data` are the same as `data` from constructor, so we don't need to pass it again explicitly

app/uk/gov/hmrc/gform/validation/DateValidation.scala
```diff
@@ -0,0 +1,340 @@
...
+import uk.gov.hmrc.gform.validation.ComponentsValidator.errors
+import uk.gov.hmrc.gform.validation.ValidationServiceHelper._
+import uk.gov.hmrc.gform.validation.ValidationUtil._
+import uk.gov.hmrc.gform.validation.ComponentsValidator._
+import uk.gov.hmrc.gform.views.html.localisation
+import uk.gov.hmrc.gform.validation.DateValidation._
+
+import scala.util.{ Failure, Success, Try }
+
+case class DateValidation(data: FormDataRecalculated) {
```
josef-vlach: All method on this class should be marked as `private` if they can be. And any method which takes `data: FormDataRecalculated` as a parameter should reuse `data` from constructor.

PR #713, opened 2019-04-09
GFC-796 - hmrcTaxPeriodSelector save and come back later behaviour

PasGat:
Just a renaming

Extract some functionality

Add returning user flag

### DIFF
--- a/app/uk/gov/hmrc/gform/models/ObligationValidator.scala
+++ b/app/uk/gov/hmrc/gform/models/ObligationValidator.scala
@@ -16,50 +16,33 @@
 
 package uk.gov.hmrc.gform.models
 
-import uk.gov.hmrc.gform.graph.Data
-import uk.gov.hmrc.gform.sharedmodel.{ Obligation, TaxResponse }
+import cats.data.NonEmptyList
+import uk.gov.hmrc.gform.sharedmodel.form.FormDataRecalculated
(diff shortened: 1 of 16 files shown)


app/uk/gov/hmrc/gform/models/ObligationValidator.scala
```diff
@@ -16,50 +16,33 @@
...
-      .size
 
-  private val goBackIf: Boolean => TaxSelectionNavigator =
-    predicate => if (predicate) GoBackToTaxPeriodSelection else DoNotGoBackToTaxPeriodSelection
+  private def mayClear(
+    formDataRecalculated: FormDataRecalculated,
+    cachedObligation: NonEmptyList[TaxResponse],
+    responseObligation: NonEmptyList[TaxResponse]): Boolean =
+    cachedObligation.toList
+      .zip(responseObligation.toList)
```
josef-vlach: `cachedObligation` and `responseObligation` are not guaranteed to be in the same order unfortunately.

The reason is that they are dependent on a dependency graph construction which is not stable. For example [here](https://github.com/hmrc/gform-frontend/blob/master/test/uk/gov/hmrc/gform/graph/DependencyGraphSpec.scala#L242) we are sorting the dependency graph explicitly for convenience in our tests.

PR #716, opened 2019-04-12
David/gfc 853/legacy fc enrolment verifier

Davidrymer97:


app/uk/gov/hmrc/gform/gform/EnrolmentController.scala
```diff
@@ -215,10 +216,18 @@ class EnrolmentController(
...
-        } yield authRes
+          _             <- validateIdentifiers[F](identifierss, postCheck)
+          _             <- enrolmentService.enrolUser(serviceId, identifiers, verifiers, retrievals)
+          initialResult <- checkEnrolment(identifiers)
+
+          reattemptResult <- (initialResult, legacyFcEnrolmentVerifier) match {
+                              case (EnrolmentFailed, Some(lfcev)) =>
+                                enrolmentService
+                                  .enrolUser(serviceId, identifiers, List(Verifier(lfcev.value, "FC")), retrievals)
+                                checkEnrolment(identifiers)
```
josef-vlach: Just a side note. This is race condition as line 225 and 227 will run concurrently since they are not part of single for comprehension. 

Let's make a function `f` which will take `verifiers` as an input and let's reuse it instead. Something like:
```
def f(verifiers: List[Verifier]) = for {
  _      <- enrolmentService.enrolUser(serviceId, identifiers, verifiers, retrievals)
  result <- checkEnrolment(identifiers)
} yield result
```

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/AuthConfig.scala
```diff
@@ -118,34 +118,49 @@ case class HmrcAgentModule(agentAccess: AgentAccess) extends AuthConfig
...
+      case _ => None
+    }
 }
 
 object AuthConfig {
 
   def toEnrolmentPostCheck(maybeRegimeId: Option[RegimeId]): EnrolmentPostCheck =
     maybeRegimeId.fold(NoCheck: EnrolmentPostCheck)(RegimeIdCheck.apply)
 
+  def enrolmentActionMatch(enrolmentAction: Option[EnrolmentAction]): EnrolmentAction = enrolmentAction match {
```
josef-vlach: Body of this method may be shortened to just `enrolmentAction.getOrElse(NoAction)`

PR #725, opened 2019-04-26
GFC-878][DC] added shortText(min,max)

DarylMConway:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/FormatExpr.scala
```diff
@@ -177,7 +177,10 @@ final case class PositiveNumber(
     extends TextConstraint
 
 case object BasicText extends TextConstraint
-case object ShortText extends TextConstraint
+case class ShortText(
+  min: Int = TextConstraint.defaultShortTextMinChars,
+  max: Int = TextConstraint.defaultShortTextMaxChars)
```
josef-vlach: Can we remove default values from here, we do not need them.

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/FormatExpr.scala
```diff
@@ -205,6 +206,8 @@ object Sterling {
 object TextConstraint {
   val defaultWholeDigits = 11
   val defaultFactionalDigits = 2
+  val defaultShortTextMinChars = 0
+  val defaultShortTextMaxChars = 1000
```
josef-vlach: These two values are not used anymore, so we can remove them.

PR #727, opened 2019-04-26
Gfc 878/text format 2

DarylMConway:


app/uk/gov/hmrc/gform/validation/DateValidationLogic.scala
```diff
@@ -127,11 +127,13 @@ object DateValidationLogic {
     result.trim
   }
 
-  def isNumeric(str: String, label: String): ValidatedNumeric =
-    Try(str.toInt) match {
-      case Success(x) => Valid(x)
-      case Failure(_) => Invalid(s"$label must be numeric")
-    }
+  def isNumeric(str: String, timeUnitLabel: String, label: String): ValidatedNumeric =
+    if (str.length == 0) Invalid(s"$label must be entered")
```
josef-vlach: `str.isEmpty` would be more idiomatic.

PR #728, opened 2019-05-09
GFC-873][DC] UK address validation errors

DarylMConway:


app/uk/gov/hmrc/gform/validation/AddressValidation.scala
```diff
@@ -89,6 +99,32 @@ object AddressValidation {
...
+            s"Building and street is longer than ${ValidationValues.addressLine} characters")).invalid
+      case (value :: Nil, Second()) if value.length > ValidationValues.addressLine =>
+        Map(
+          fieldId -> errors(
+            fieldValue,
+            s"Building and street line 2 is longer than ${ValidationValues.addressLine} characters")).invalid
+      case (value :: Nil, Third()) if value.length > ValidationValues.addressLine =>
+        Map(fieldId -> errors(fieldValue, s"Town or city is longer than ${ValidationValues.addressLine} characters")).invalid
+      case (value :: Nil, Fourth()) if value.length > ValidationValues.addressLine4 =>
+        Map(fieldId -> errors(fieldValue, s"County is longer than ${ValidationValues.addressLine4} characters")).invalid
```
josef-vlach: Let's create method like this one 
```
def abc(str: String) = Map(fieldId -> errors(fieldValue, str).invalid
```
and use it for these four matches

PR #746, opened 2019-05-31
GFC-715 sign out controller

nubz:


app/uk/gov/hmrc/gform/views/main_template.scala.html
```diff
@@ -34,7 +34,8 @@
   frontendAppConfig: FrontendAppConfig,
   authConfig: Option[AuthConfig],
   webChat: Option[WebChat] = None,
-  formCategory: FormCategory = Default
+  formCategory: FormCategory = Default,
+  templateId: String
```
josef-vlach: Let's keep type information here ie. use `FormTemplateId` instead of `String`

app/uk/gov/hmrc/gform/views/govuk_wrapper.scala.html
```diff
@@ -37,7 +37,8 @@
   scriptElem: Option[Html] = None,
   authConfig: Option[AuthConfig],
   navLinks: Option[Html] = None,
-  includeHMRCBranding: Boolean
+  includeHMRCBranding: Boolean,
+  templateId: String
```
josef-vlach: Let's keep type information here ie. use `FormTemplateId` instead of `String`

app/uk/gov/hmrc/gform/views/main_template.scala.html
nubz: Done

app/uk/gov/hmrc/gform/views/govuk_wrapper.scala.html
nubz: Done

PR #764, opened 2019-06-12
[GFC-901][DC] defect: form name still string

DarylMConway:


test/uk/gov/hmrc/gform/sharedmodel/formtemplate/generators/FormTemplateGen.scala
```diff
@@ -51,6 +51,8 @@ trait FormTemplateGen {
       templateName <- templateNameGen
     } yield WebChat(ChatRoomId(roomId), templateName)
 
+  def toLocalisedString(s: String): LocalisedString = LocalisedString(Map(LangADT.En -> s))
```
josef-vlach: This function already exists, let's reuse existing one.

DarylMConway: swapped it use the helpers toLocalisedString

PR #769, opened 2019-06-18
GFC-932: removed wrapper method for validateComponents

nubz:


app/uk/gov/hmrc/gform/gform/DeclarationController.scala
```diff
@@ -108,23 +108,18 @@ class DeclarationController(
...
-      valRes                            <- validateComponents(cacheWithHiddenSectionDataRemoved, declarationData)
-      response                          <- processValidation(valRes, maybeAccessCode, cacheWithHiddenSectionDataRemoved, declarationData)
+      valRes <- validationService.validateComponents(
+                 Fields.flattenGroups(cacheWithHiddenSectionDataRemoved.formTemplate.declarationSection.fields),
+                 declarationData,
+                 cacheWithHiddenSectionDataRemoved.form.envelopeId,
+                 cacheWithHiddenSectionDataRemoved.retrievals,
+                 cacheWithHiddenSectionDataRemoved.form.thirdPartyData,
+                 cacheWithHiddenSectionDataRemoved.formTemplate
+               )
```
PasGat: looks good - I would only extract  validationService.validateComponents in to a different function something like
validationService.validateComponents(cacheWithHiddenSectionDataRemoved) that will also safe you from doing cacheWithHiddenSectionDataRemoved.this cacheWithHiddenSectionDataRemoved.that

nubz: Created new method for handling cache object instead of params list, let me know if this works for you.

PR #789, opened 2019-07-11
GFC-947: add upscan-initiate connector & service

co-category:


josef-vlach:
I believe we can close this since `upscan` integration has been scrapped.

co-category:
Yes, we don't need this for now.

PR #794, opened 2019-07-23
MOR-107: allow specification of admin only controller methods

co-category:


app/uk/gov/hmrc/gform/auth/AuthService.scala
```diff
@@ -79,7 +80,8 @@ class AuthService(
   private val notAuthorized: AuthResult = AuthBlocked("You are not authorized to access this service")
   private val decoder = Base64.getDecoder
 
-  private def performAWSALBAuth(assumedIdentity: Option[Cookie])(implicit hc: HeaderCarrier): AuthResult = {
+  private def performAWSALBAuth(assumedIdentity: Option[Cookie], mustBeAdmin: Boolean)(
```
josef-vlach: So `mustBeAdmin: Boolean` parameter is basically *render layer* telling *authorisation layer* how it should do the authorisation.

It should really be *render layer* depending on result of *authorisation layer*.

Approach with `mustBeAdmin: Boolean` won't really scale and it is basically tangling both layers together. What this solution is missing is any kind of abstraction in terms of which we would decide which page/action is accessible and which is not.

This missing abstraction is what @lancewalton is currently working on, so I suggest we should implement this ticket based on his work once it is done.

co-category: Yes I agree. Keep this PR alive for his reference. We discussed this yesterday.

co-category: Once @lancewalton  has done it we can trash this.

PR #796, opened 2019-07-29
MOR-41: User can only perform functions appropriate to their role and…

lancewalton:
… the form status

### DIFF
--- a/app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
+++ b/app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
@@ -298,7 +298,7 @@ class AuthenticatedRequestActions(
   }
 
   val defaultRetrievals = Retrievals.authProviderId and Retrievals.allEnrolments and
-    Retrievals.affinityGroup and Retrievals.internalId and
+    Retrievals.internalId and
     Retrievals.externalId and Retrievals.userDetailsUri and
     Retrievals.credentialStrength and Retrievals.agentCode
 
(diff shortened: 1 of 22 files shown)


app/uk/gov/hmrc/gform/gform/FormController.scala
```diff
@@ -249,93 +258,68 @@ class FormController(
...
+                continue_form_page(
+                  cache.formTemplate,
+                  choice.bindFromRequest().withError("decision", "error.required"),
+                  maybeAccessCode,
+                  frontendAppConfig)).pure[Future], {
+              case "continue" =>
+                val dataRaw = FormDataHelpers.formDataMap(cache.form.formData) + cache.form.visitsIndex.toVisitsTuple
+                redirectWithRecalculation(cache, dataRaw, maybeAccessCode, ForceReload)
+              case "delete" =>
+                // ToDo: Lance - could we make this redirect to delete(., .) to make use of the permissions checking there?
```
josef-vlach: I guess this comment is not needed anymore, now that delete require the same permission?

app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
```diff
@@ -281,47 +316,48 @@ class AuthenticatedRequestActions(
 
     authorised(predicate)
       .retrieve(defaultRetrievals) {
-        case authProviderId ~ enrolments ~ affinityGroup ~ internalId ~ externalId ~ userDetailsUri ~ credentialStrength ~ agentCode =>
+        case authProviderId ~ enrolments ~ _ ~ internalId ~ externalId ~ userDetailsUri ~ credentialStrength ~ agentCode =>
```
josef-vlach: instead of ignoring affinity group like this `~ _ ~` it would be better not to ask for it in the first place by modifying `defaultRetrievals`

app/uk/gov/hmrc/gform/gform/FormController.scala
lancewalton: Removed

app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
lancewalton: Sorted

PR #800, opened 2019-08-05
GFC-970 - Implement draft retrieval method for all users (formAccessCode, as per spec) - Ofsted

josef-vlach:


test/uk/gov/hmrc/gform/gformbackend/DemoApp.scala
```diff
@@ -39,9 +39,9 @@ object DemoApp extends App {
   val userId = UserId("TestUser")
 
   val result = for {
-    formId <- gformConnector.newForm(formTemplateId, userId, None)
-    form   <- gformConnector.getForm(formId)
-    _ = println(s"formId = $formId")
+    newFormData <- gformConnector.newForm(formTemplateId, userId, None)
+    form        <- gformConnector.getForm(newFormData.formId)
+    _ = println(s"formId = ${newFormData.formId}")
```
lancewalton: Are you deliberately leaving these printlns in?

josef-vlach: I'm not sure what purpose of `DemoApp` is. We probably don't need it, but since println was there before I simply do not want to remove it, in case someone relies on it 🤷‍♂.

PR #812, opened 2019-08-27
GFC-991: When a form is submitted, wipe down data for all form fields…

lancewalton:
… in hidden sections


lancewalton:
OK. I've tried a different approach...

PR #822, opened 2019-09-17
GFC-1081: Includeifs to work with multivalue

lancewalton:
I'm not sure about this implementation. In particular BooleanExprEval.includes

Unrelated we should probably move BooleanExprEval out of BooleanExpr.scala & out of sharedmodel.

### DIFF
--- a/app/uk/gov/hmrc/gform/graph/Recalculation.scala
+++ b/app/uk/gov/hmrc/gform/graph/Recalculation.scala
@@ -30,9 +30,10 @@ import scalax.collection.Graph
 import scalax.collection.GraphEdge._
 import uk.gov.hmrc.gform.auth.models.MaterialisedRetrievals
 import uk.gov.hmrc.gform.commons.{ BigDecimalUtil, NumberFormatUtil }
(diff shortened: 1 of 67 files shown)


app/uk/gov/hmrc/gform/graph/Recalculation.scala
```diff
@@ -257,7 +258,11 @@ class Recalculation[F[_]: Monad, E](
     }
   }
 
-  private def hasData(fc: FormComponent, dataLookup: Data): Boolean =
+  // TODO: Is this correct? The logic is:
+  // a) data exists for fc, &&
+  // b) there are no empty values in the Seq[String]
+  // I would have thought (b) should be "there is at least one String in the sequence which is non-empty
```
josef-vlach: This works first (b) since only `HmrcTaxPeriod`, `Text` and `TextArea` are relevant and they have Seq of size one. New (b) would be ok as well.

In other words `Seq[String]` is going to be more then one element only for checkboxes and they are irrelevant, because there is no expression to be reevaluated associated with them.

PR #844, opened 2019-11-06
Updating generic error message UK EORI

Jonathan-Courage:


Jonathan-Courage:
Was it just this line that needed updated? 

Cheers

Jonathan-Courage:
Added in Welsh TODO from Josef feedback.

PR #872, opened 2020-01-22
Adding correct spacing/ array to csv files and lookupLoader

Jonathan-Courage:


app/uk/gov/hmrc/gform/LookupLoader.scala
```diff
@@ -69,23 +69,23 @@ class LookupLoader {
   private def mkRadioLookup(m: LocalisedLookupOptions): RadioLookup = RadioLookup(m)
 
   // format: off
-  private val cashType                 = read("BCD-CashType.csv",                 "ID",           "en",   "cy",   mkRadioLookup)
```
josef-vlach: It would be nice to keep column alignment in this block of code. So instead of all these whitespace changes only indent `currency` and `country` by two spaces.

Jonathan-Courage: Screenshot from 2020-01-22 15-31-09

Like that, so it all lines up?

josef-vlach: Yes 👍

Jonathan-Courage: Fixed in latest commit :)

Jonathan-Courage:
Tested in QA with Ratheesh. Works.

PR #933, opened 2020-08-04
GFORMS-847 - Accessibility error fixes

SandipKhedekar:


app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -712,7 +712,7 @@ class SectionRenderingService(frontendAppConfig: FrontendAppConfig, lookupRegist
     )
 
     val fileUpload: fileupload.FileUpload = fileupload.FileUpload(
-      id = formComponent.id.value,
+      id = formComponent.id.value + "0",
```
josef-vlach, line 715: Why are you adding this `"0"`?

PR #959, opened 2020-09-28
GFORMS-734 - Solve problem of what data is visible and invisible in t…

josef-vlach:
…he form model


girishkamat:
@josef-vlach There is a conflict that needs resolving

PR #963, opened 2020-09-30
GFORMS-925 - Remove section heading from browser page title

SandipKhedekar:


app/uk/gov/hmrc/gform/views/form/form.scala.html
```diff
@@ -68,7 +68,7 @@
 @govukButton = @{new govukButton()}
 
 @main_template(
-  title = heading,
+  title = if(isDeclaration) heading else "",
```
josef-vlach, line 71: I don't think we want declaration to be treat differently, since declaration title is also `SmartString` so it can contains PII.  So `title` in here should be always empty `""`

app/uk/gov/hmrc/gform/views/hardcoded/pages/save_with_access_code.scala.html
```diff
@@ -24,7 +24,7 @@
 )(implicit request: Request[_], messages: Messages, l: LangADT)
 
 @uk.gov.hmrc.gform.views.html.main_template(
-  title = saveWithAccessCode.heading,
+  title = "",
```
josef-vlach, line 27: `saveWithAccessCode.heading` seems to be harmless, since it originate in messages.

app/uk/gov/hmrc/gform/views/hardcoded/pages/start_new_form.scala.html
```diff
@@ -33,7 +33,7 @@
 @govukButton = @{new govukButton()}
 @button = @{messages("button.continue")}
 
-@heading = @{messages("accessCode.new.title", displayAccessCode.accessCodeName)}
+@heading = @{messages("accessCode.new.title")}
```
josef-vlach, line 36: `displayAccessCode.accessCodeName` seems to be used in:

```
accessCode.new.title=Make a note of your {0}
```

josef-vlach (changes requested):

josef-vlach (approved):

PR #966, opened 2020-10-06
GFORMS-937 - The MTD VAT and Vishing form need numbers entering that …

josef-vlach:
…should not display as a number

### DIFF
--- a/app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
+++ b/app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
@@ -110,6 +110,7 @@ object ComponentValidator {
           ValidationValues.sterlingLength,
           TextConstraint.defaultFactionalDigits,
           s.positiveOnly)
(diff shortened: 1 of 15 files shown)


app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
```diff
@@ -251,6 +252,17 @@ object ComponentValidator {
...
   }
 
+  private def referenceNumberConstraints(fieldValue: FormComponent, value: String, min: Int, max: Int)(
+    implicit
+    messages: Messages,
+    sse: SmartStringEvaluator
+  ) = {
+    val ValidReferenceNumber = s"[0-9]{$min,$max}".r
+    val str = value.replace(" ", "")
+    val messageKey = "generic.referenceNumber.error.pattern"
```
girishkamat, line 262: would it be useful to also include min and max length allowed, in the error message?

josef-vlach: Min/Max length is dealt with in the `sharedTextComponentValidator` method.

girishkamat (approved):

PR #969, opened 2020-10-09
GFORM-916 Refactor

girishkamat:


app/uk/gov/hmrc/gform/lookup/LookupType.scala
```diff
@@ -21,9 +21,9 @@ import uk.gov.hmrc.gform.sharedmodel.LangADT
 
 sealed trait LookupType extends Product with Serializable
 
-case class RadioLookup(options: LocalisedLookupOptions) extends LookupType
-case class AjaxLookup(
-  options: LocalisedLookupOptions,
+case class RadioLookup[T](options: LocalisedLookupOptions[T]) extends LookupType
```
josef-vlach, line 24: This introduction of existential type is something I'd rather avoid. It is causing problems down the line.
Check compiler warnings to see what I mean:

```
[warn]  [E1] app/uk/gov/hmrc/gform/gform/StructuredFormDataBuilder.scala:173:16:
[warn]       match may not be exhaustive.
[warn]       It would fail on the following input: Some((x: Any forSome x not in (uk.gov.hmrc.gform.lookup.CountryLookupInfo, uk.gov.hmrc.gform.lookup.CurrencyLookupInfo, uk.gov.hmrc.gform.lookup.DefaultLookupInfo, uk.gov.hmrc.gform.lookup.PortLookupInfo)))
[warn]             .flatMap {
```

PR #989, opened 2020-11-10
GFORMS-927 Fixed issues with display of add to list sections

girishkamat:


app/uk/gov/hmrc/gform/instructions/InstructionsRenderingService.scala
```diff
@@ -210,41 +214,82 @@ class InstructionsRenderingService(
...
-            section.fold(_ => List.empty[HtmlFormat.Appendable])(_ => List.empty[HtmlFormat.Appendable])(addToList =>
-              addToList.instruction.map(i => begin_section(i.name)).toList)
-          header ++ pageModels.flatMap(pModel => {
-            pModel.fold(renderHtmls(_, formModel.pagesMap(pModel)))(_ => Nil)
-          })
+        case (a: AddToList, pagesWithSectionNumber) =>
+          List(addToListRender(a)) ++ pagesWithSectionNumber
+            .collect {
+              case (singleton: Singleton[Visibility], sectionNumber) =>
+                (formModel.repeaterForSingleton(singleton, sectionNumber).get, (singleton, sectionNumber))
```
josef-vlach, line 253: Have you consider using existing
```
def repeaterFor(addToListId: AddToListId): Option[Repeater[A]]
```
here ?

Also please don't call `.get` on an `Option`

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/instructions/InstructionsRenderingService.scala
girishkamat: Thanks @josef-vlach. I have removed the usage of Option.get

I found the logic in SummaryRenderingService, using `repeaterFor(index: Int, addToListId: AddToListId)` a bit hard to follow, thats why resorted to using grouping. To use `repeaterFor(index: Int, addToListId: AddToListId)`, I need to remember the previous repeaters index, which I don't have here. Do you see an issue with this approach?

josef-vlach (approved):

PR #992, opened 2020-11-16
GFORMS-959 Added support for summary heading in add to list section

girishkamat:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Section.scala
```diff
@@ -96,6 +96,7 @@ object Section {
     title: SmartString,
     description: SmartString,
     shortName: SmartString,
+    summaryName: Option[SmartString],
```
josef-vlach, line 99: Shouldn't this be mandatory?

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Section.scala
girishkamat: I have left it as Optional, so it works for existing form templates already saved in database.

josef-vlach: We don't need to keep backward functionality. That is why we have "Reload All" functionality in Admin tool.

girishkamat: Cool, will make the change then

josef-vlach (approved):

PR #994, opened 2020-11-17
GFORMS-927 Fixed issue with page title for PDF documents

girishkamat:


app/uk/gov/hmrc/gform/summary/SummaryRenderingService.scala
```diff
@@ -90,7 +90,11 @@ class SummaryRenderingService(
         HtmlSanitiser
           .sanitiseHtmlForPDF(
             summaryHtml,
-            document => HtmlSanitiser.acknowledgementPdf(document, submissionDetailsString, cache.formTemplate)))
+            document => {
+              document.title(s"Form Summary - ${cache.formTemplate.formName.value}")
```
josef-vlach, line 94: `Form Summary` shouldn't this be localised?

girishkamat: Probably yes, thanks

app/uk/gov/hmrc/gform/summary/FormComponentSummaryRenderer.scala
```diff
@@ -0,0 +1,924 @@
...
+  )(
+    implicit
+    messages: Messages,
+    lise: SmartStringEvaluator,
+    fcrd: FormComponentRenderDetails[T]): List[SummaryListRow] = {
+
+    val hasErrors = formFieldValidationResult.isNotOk
+
+    val errors = checkErrors(formComponent, formFieldValidationResult)
+    val label = fcrd.label(formComponent).capitalize
```
josef-vlach, line 498: This `capitalize` is now acting differently than before. But I'm not sure why it is there in the first place.

app/uk/gov/hmrc/gform/summary/FormComponentSummaryRenderer.scala
```diff
@@ -0,0 +1,924 @@
...
+}
+
+sealed trait RenderType
+trait InstructionRender extends RenderType
+trait SummaryRender extends RenderType
+
+sealed trait FormComponentRenderDetails[T <: RenderType] {
+  def isVisible(formComponent: FormComponent): Boolean
+  def label(formComponent: FormComponent)(implicit lise: SmartStringEvaluator): String
+  def sort(fields: List[FormComponent]): List[FormComponent]
```
josef-vlach, line 895: Cannot we merge `isVisible` and `sort` into single method? They seems to be used exclusively in combination.

app/uk/gov/hmrc/gform/summary/FormComponentSummaryRenderer.scala
girishkamat: Thats correct, however I dont think it will make a lot of difference and not sure why its even needed

app/uk/gov/hmrc/gform/summary/FormComponentSummaryRenderer.scala
girishkamat: I thought so too, but the name would be a bit wierd. sortAndGetRenderables?

josef-vlach: That `sort` should not be even exposed in the API. It is just detail in `InstructionRender` instance.
I think it would be best to have 

```
sealed trait FormComponentRenderDetails[T <: RenderType] {
  def prepareRenderables(fields: List[FormComponent]): List[FormComponent]
  def label(formComponent: FormComponent)(implicit lise: SmartStringEvaluator): String
}
```

girishkamat: Excellent suggestion!

josef-vlach (approved):

PR #1009, opened 2020-11-25
GFORMS-814 Added flag to show/hide HMRC logo on all pages and pdfs

girishkamat:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/FormCategory.scala
```diff
@@ -20,7 +20,6 @@ import play.api.libs.json._
 
 sealed trait FormCategory {
   def getString: String
-  def isHmrc: Boolean = this != OfstedDefault
```
josef-vlach, line 23: So is the plan to replace `"formCategory": "ofstedDefault",`
with this new flag in all our forms?

girishkamat: Yes, the ticket mentions that.

girishkamat: I think the formCategory will remain. however it wont be used to hide/display  hmrc logo

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/FormTemplate.scala
```diff
@@ -37,9 +37,10 @@ case class FormTemplate(
   parentFormSubmissionRefs: List[FormComponentId],
   languages: AvailableLanguages,
   save4LaterInfoText: Option[Save4LaterInfoText],
-  summarySection: SummarySection
-) {
+  summarySection: SummarySection,
+  displayHMRCLogo: Option[Boolean]) {
```
josef-vlach, line 41: Can't we this make just a `Boolean`?

girishkamat: Option[Boolean] reflects the fact that this is indeed an optional value in the template. I was trying to make a simple Boolean, by introducing something similar to  UploadableHmrcDestination, however felt its better to leave it optional. Is there any other way other than using a complementary case class to convert to the main class?

josef-vlach: We don't need to reflect the fact how things are defined in template, if it is not good for our code.

If we want to see template differently (for example as if `displayHMRCLogo` was mandatory) we are preprocessing template in `normaliseJSON` [see here](https://github.com/hmrc/gform/blob/master/app/uk/gov/hmrc/gform/formtemplate/FormTemplatesControllerRequestHandler.scala#L79-L81). With this approach we don't need any complementary case class to achieve that.

girishkamat: Thanks Alex, thats a valuable suggestion. I have updated the code accordingly

josef-vlach (approved):

PR #1013, opened 2020-11-27
GFORMS-913 - Turn every empty StringResult into Empty

josef-vlach:


app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -127,8 +127,11 @@ case class EvaluationResults(
     evaluationContext: EvaluationContext
   ): ExpressionResult = {
 
+    def nonEmpty(stringResult: StringResult): ExpressionResult =
```
girishkamat, line 130: may be call it defaultToEmpty?

girishkamat (approved):

PR #1017, opened 2020-11-30
GFORMS-990 Updated heading font to govuk-heading-l in AddToList summary page

girishkamat:


app/uk/gov/hmrc/gform/views/form/addToList.scala.html
```diff
@@ -79,8 +79,9 @@
 
     @pageLevelError.render
 
-    @hmrcPageHeading(PageHeading(text = heading))
```
josef-vlach, line 82: If we don't use `hmrcPageHeading` value anymore, let's clean it up and remove import

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/views/form/addToList.scala.html
girishkamat: Good point, done

app/uk/gov/hmrc/gform/views/form/addToList.scala.html
```diff
@@ -79,8 +77,9 @@
...
-
+    <h1 class="govuk-heading-l">
```
josef-vlach, line 80: Also let's use our own `pageHeading.scala.html` for consistency as we do elsewhere instead of custom html.

girishkamat: Done

josef-vlach (approved):

PR #1022, opened 2020-12-03
GFORMS-989 Fixed font size issue with delete link in uploaded file list

girishkamat:


app/uk/gov/hmrc/gform/views/govuk_wrapper.scala.html
```diff
@@ -176,7 +176,8 @@
     @hmrcReportTechnicalIssue(
       ReportTechnicalIssue(
         serviceCode = appConfig.contactFormServiceIdentifier,
-        language = Some(language.code)
+        language = Some(language.code),
+        classList = Some("govuk-body")
```
josef-vlach, line 180: I'm not sure this is correct. Issue is about delete link in file-upload component.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/views/govuk_wrapper.scala.html
girishkamat: The font for Delete link is coming form govuk-body class and that applies to all the elements in the body. However, the report technical issues link is outside the body, while it should have been inside. Putting it inside means copying it into every template and deleting it from the govuk_template. This is a quick fix to make all links consistent

girishkamat: @josef-vlach I have added the govuk-link to the Delete anchor. Please take a look

girishkamat: Had a chat with @richardhattonUX on the changes and he seems happy with it

josef-vlach: @girishkamat Look's good, but can we remove `delete-file` class, it seems have no effect.

girishkamat: @josef-vlach It looks like some javascript code uses it to lookup the elements

`    function init () {
      $('.govuk-file-upload').on('change', handleFileUpload);
      $('.uploaded-files').on('click', '.delete-file', handleFileDelete);
    }`

josef-vlach (approved):

PR #1038, opened 2020-12-30
Gforms 890 play upgrade 2.7

girishkamat:


app/uk/gov/hmrc/gform/gform/GovukExtensions.scala
```diff
@@ -0,0 +1 @@
+
```
josef-vlach, line 1: Shouldn't this file be removed?

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/gform/GovukExtensions.scala
girishkamat: I dont see this file in my branch. When I rebased from master, this did show up as deleted

girishkamat: For some reason search didnt show it. I have removed it now

josef-vlach (approved):

PR #1044, opened 2021-01-13
GFORMS-712 Added support for date comparison in includeIf

girishkamat:


app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -178,6 +182,36 @@ case class EvaluationResults(
...
+    def loop(expr: Expr): ExpressionResult = expr match {
+      case ctx @ FormCtx(formComponentId) =>
+        val exprResult: ExpressionResult = get(ctx, recData, fromVariadicValue)
+        exprResult.fold[ExpressionResult](identity)(_ => exprResult)(_ => {
+          val year = get(formComponentId.toAtomicFormComponentId(Atom("year")), recData)
+          val month = get(formComponentId.toAtomicFormComponentId(Atom("month")), recData)
+          val day = get(formComponentId.toAtomicFormComponentId(Atom("day")), recData)
+          (year, month, day) match {
+            case (Some(VariadicValue.One(y)), Some(VariadicValue.One(m)), Some(VariadicValue.One(d))) =>
+              StringResult(y.toInt.formatted("%04d") + m.toInt.formatted("%02d") + d.toInt.formatted("%02d"))
```
josef-vlach, line 206: Why not to store it here as `DateResult`?

app/uk/gov/hmrc/gform/models/FormModelBuilder.scala
```diff
@@ -104,6 +97,9 @@ class FormModelBuilder[E, F[_]: Functor](
       case ExpressionResult.StringResult(value) => VariadicFormData.one[SourceOrigin.Current](modelComponentId, value)
       case ExpressionResult.OptionResult(value) =>
         VariadicFormData.many[SourceOrigin.Current](modelComponentId, value.map(_.toString))
+      case ExpressionResult.DateResult(_) =>
+        VariadicFormData
+          .empty[SourceOrigin.Current] //TODO: geneate DateResult instead of StringResult for ExprType DateString
```
josef-vlach, line 102: This implementation looks suspicious. Why `empty`? Is it relate to previous comment?

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -51,13 +52,16 @@ case class EvaluationResults(
...
     fromVariadicValue: VariadicValue => ExpressionResult
-  ): ExpressionResult =
-    exprMap
-      .get(expr)
-      .getOrElse(
-        recData.variadicFormData
-          .get(expr.formComponentId.modelComponentId)
-          .fold(ExpressionResult.empty)(fromVariadicValue))
+  ): ExpressionResult = {
+    val result = recData.variadicFormData
```
josef-vlach, line 56: This result may not be needed, why to compute it all the time?

josef-vlach (changes requested):
Looks good, only couple of minor issue in the comments.

But in total it doesn't work correctly for more complex jsons. If there is an expression which depends on field on includeIf (with date) section, that expression don't work correctly.

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
girishkamat: Yes, had the same idea. Will do the change

app/uk/gov/hmrc/gform/models/FormModelBuilder.scala
girishkamat: correct. How to represent LocalDate as string as VariadicValue only supports String right?

josef-vlach: oh, yes, we don't want to put `DateResult` into `VariadicFormData`, so `empty` is correct

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
girishkamat: Accidental commit. changed for debugging but forgot to remove.

app/uk/gov/hmrc/gform/models/FormModelBuilder.scala
girishkamat: I have modified to put a string representation of Date into VariadicFormData (yyyymmdd). That should be ok I guess?

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
girishkamat: done

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
girishkamat: removed

girishkamat:
Lets have a discussion to walk through the changes. Although I have tested and it seems to work, I am not very confident of the changes, I may have unintenationally missed something and introduced a bug somewhere

josef-vlach (approved):

PR #1045, opened 2021-01-15
GFORMS-974 - Add new expression to check a value against a regular ex…

josef-vlach:
…pression

### DIFF
--- a/app/uk/gov/hmrc/gform/graph/Recalculation.scala
+++ b/app/uk/gov/hmrc/gform/graph/Recalculation.scala
@@ -21,6 +21,7 @@ import cats.syntax.all._
 import cats.data.StateT
 
 import scala.language.higherKinds
+import scala.util.matching.Regex
 import scalax.collection.Graph
 import scalax.collection.GraphEdge._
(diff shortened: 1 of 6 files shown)


app/uk/gov/hmrc/gform/graph/Recalculation.scala
```diff
@@ -88,6 +89,14 @@ class Recalculation[F[_]: Monad, E](
         noStateChange(res)
       }
 
+      def matchRegex(formCtx: FormCtx, regex: Regex): Boolean = {
+        val typeInfo1 = formModel.toFirstOperandTypeInfo(formCtx)
```
girishkamat, line 93: As we know we are matching RegEx, which is a string. Do we need to determined type first? Can we just convert whatever the value is to string and evaluate?

josef-vlach: We can match numbers as well. But more importantly we need to utilise the heavy work done by `ExpressionResult` since in `includeIf` we may be matching against invisible field.

girishkamat (approved):

PR #1050, opened 2021-01-21
GFORMS-940 - Investigate data sent to Google Analytics to include Cho…

josef-vlach:
…ice label rather than index


app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -1703,6 +1710,11 @@ class SectionRenderingService(frontendAppConfig: FrontendAppConfig, lookupRegist
     }
   }
 
+  private def dataLabelAttribute(label: SmartString): Map[String, String] =
+    dataLabelAttribute(label.localised.value(LangADT.En))
+  private def dataLabelAttribute(label: String): Map[String, String] =
+    Map("data-label" -> label.replaceAll("''", "'"))
```
girishkamat, line 1716: What is the need to replace quotes like this?

josef-vlach: To unescape `'`:
https://tonnygaric.com/blog/escape-single-quotes-when-using-java-messageformat

girishkamat (approved):

PR #1054, opened 2021-01-26
GFORMS-1070 - Prevent the internal field id from being displayed when…

josef-vlach:
… file name used in smart tex


SandipKhedekar (approved):

josef-vlach:
Fixed by #1061

PR #1056, opened 2021-01-29
GFORMS-1061 - Add to List Back link changes

SandipKhedekar:


josef-vlach (changes requested):
Looks neat, but little more work is needed. We need to handle gracefully the case when add-to-list is very first section on the form. Currently back link leading nowhere is present.

josef-vlach (approved):

PR #1060, opened 2021-02-04
GFORMS-1077 Support Else expression using dates

girishkamat:


app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -196,6 +196,8 @@ case class EvaluationResults(
     def loop(expr: Expr): ExpressionResult = expr match {
       case ctx @ FormCtx(_) =>
         evalDateExpr(recData, this)(DateFormCtxVar(ctx))
+      case Else(field1: Expr, field2: Expr) =>
+        loop(field1) orElse loop(field2)
```
josef-vlach, line 200: This looks weird, will this work for expression which mixes date and other types?

Like:

```
${startDate else 'You didnt select any date'}
```

girishkamat: Thats a good point, haven't tested it. It will return Empty for non date values (when condition is met)

girishkamat: Just added some changes. Basically if the date expression could not be evaluated i.e Empty, it tries to evaluate as a String as a fallback

josef-vlach (approved):

PR #1066, opened 2021-02-10
Gforms 1083 instructionspdf tabular format

girishkamat:


app/uk/gov/hmrc/gform/instructions/PageFieldConverter.scala
```diff
@@ -0,0 +1,43 @@
...
+import play.api.i18n.Messages
+import uk.gov.hmrc.gform.controllers.AuthCacheWithForm
+import uk.gov.hmrc.gform.eval.smartstring.SmartStringEvaluator
+import uk.gov.hmrc.gform.fileupload.Envelope
+import uk.gov.hmrc.gform.instructions.FormModelInstructionSummaryConverter.PageField
+import uk.gov.hmrc.gform.sharedmodel.LangADT
+import uk.gov.hmrc.gform.sharedmodel.formtemplate.{ ComponentType, FormComponent, SectionNumber }
+import uk.gov.hmrc.gform.validation.ValidationResult
+
+trait PageFieldConverter[T <: ComponentType] {
```
josef-vlach, line 28: Is this trait really necessary?

As I see is that it forces you to do unsafe `asInstanceOf` cast, and pass almost a dozen of arguments to the functions which do not need them. 

Why don't use `IsText` and other extractors to stay consistent with the rest of the code base?

girishkamat: The idea was to have separate implementations of summary converters for each scenario - i.e InstructionPDF now, summary and summary PDF later. Also adding a new converter for a new type doesnt touch the main code, just an additional instance required. However, I do agree the long list of arguments and the casting is a bit annoying

girishkamat: @josef-vlach I have refactored the code. Please take a look

josef-vlach (approved):

PR #1083, opened 2021-02-22
GFORMS-383 - Introduce IncludeIf functionality at field level

josef-vlach:


app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -612,48 +610,58 @@ class SectionRenderingService(frontendAppConfig: FrontendAppConfig, lookupRegist
...
-              validationResult,
-              ei,
-              obligations)
-          case FileUpload() =>
-            htmlForFileUpload(formComponent, formTemplateId, ei, ei.retrievals, validationResult)
-          case InformationMessage(infoType, infoText) =>
-            htmlForInformationMessage(formComponent, infoType, infoText, ei)
-          case htp @ HmrcTaxPeriod(idType, idNumber, regimeType) =>
-            htmlForHmrcTaxPeriod(formComponent, ei, validationResult, obligations, htp)
+        val isVisible = formComponent.includeIf.fold(true) { includeIf =>
```
girishkamat, line 613: why is the evaluation needed here? By the time we are here, the renderPageOptics would have already filtered invisible form components based on the includeIf evaluation right?

test/uk/gov/hmrc/gform/models/FormModelBuilderSpec.scala
```diff
@@ -0,0 +1,97 @@
...
+        val visibilityOptics: FormModelVisibilityOptics[DataOrigin.Mongo] =
+          fmb.visibilityModel[DataOrigin.Mongo, SectionSelectorType.Normal](data, None)
+
+        val formModelOptics: FormModelOptics[DataOrigin.Mongo] =
+          fmb.renderPageModel[DataOrigin.Mongo, SectionSelectorType.Normal](visibilityOptics, None)
+
+        val expected: FormModel[Visibility] = fromPagesWithIndex(expectedPages, staticTypeInfo)
+
+        formModelOptics.formModelVisibilityOptics.formModel shouldBe expected
+        visibilityOptics.recData.variadicFormData shouldBe expectedData
```
girishkamat, line 93: do we really need the second assertion here?

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
josef-vlach: No, `renderPageOptics` contains all the components. It is `formModelVisibilityOptics` which contain only visible ones.

The reason `renderPageOptics` contains all components, is that we need to know what components are on the page, so they get rendered accordingly.

test/uk/gov/hmrc/gform/models/FormModelBuilderSpec.scala
josef-vlach: It is there as a sanity check that invisibility of the component doesn't affect data we store for such a component.

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
girishkamat: Got it, thanks

test/uk/gov/hmrc/gform/models/FormModelBuilderSpec.scala
girishkamat: Ok

girishkamat (approved):

PR #1084, opened 2021-02-23
GFORMS-1097 Fix issue with addition of add-to-list count expressions …

girishkamat:
…in value field of text components

### DIFF
--- a/app/uk/gov/hmrc/gform/models/FormModel.scala
+++ b/app/uk/gov/hmrc/gform/models/FormModel.scala
@@ -146,8 +146,9 @@ case class FormModel[A <: PageMode](
     first.fold(illegal) {
       case FormCtx(formComponentId) => explicitTypedExpr(expr, formComponentId)
       case DateCtx(_)               => TypeInfo(expr, StaticTypeData(ExprType.dateString, None))
-      case IsNumberConstant(_)      => TypeInfo(expr, StaticTypeData(ExprType.number, Some(Number())))
-      case otherwise                => TypeInfo(expr, StaticTypeData(ExprType.string, None))
+      case IsNumberConstant(d) =>
+        TypeInfo(expr, StaticTypeData(ExprType.number, Some(Number(maxFractionalDigits = d.scale))))
+      case otherwise => TypeInfo(expr, StaticTypeData(ExprType.string, None))
     }
   }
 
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/gform/models/FormModelBuilder.scala
```diff
@@ -278,7 +278,8 @@ class FormModelBuilder[E, F[_]: Functor](
     }
 
     val visibleTypedExprs: List[(FormComponentId, TypeInfo)] = visibilityFormModel.allFormComponents.collect {
-      case fc @ HasValueExpr(expr) if !fc.editable => (fc.id, visibilityFormModel.explicitTypedExpr(expr, fc.id))
+      case fc @ HasValueExpr(expr) if !fc.editable =>
+        (fc.id, visibilityFormModel.toFirstOperandTypeInfo(expr))
```
josef-vlach, line 282: I'm afraid this is wrong. `TypeInfo` for component with value expression needs to be derived from component, not from the expression.

For example type of expression `${1 + 2}` is `String`, but type of expression `${4 + 5}` is `Number`:

```json
{
    "id": "foo",
    "type":"text",
    "format": "shortText",
    "value": "${1 + 2}",
    "label": "Foo ${4 + 5}"
}
```

Method `toFirstOperandTypeInfo` is supposed to derive expression type from free expressions, those are expression in labels etc.

app/uk/gov/hmrc/gform/models/FormModel.scala
```diff
@@ -146,8 +146,9 @@ case class FormModel[A <: PageMode](
     first.fold(illegal) {
       case FormCtx(formComponentId) => explicitTypedExpr(expr, formComponentId)
       case DateCtx(_)               => TypeInfo(expr, StaticTypeData(ExprType.dateString, None))
-      case IsNumberConstant(_)      => TypeInfo(expr, StaticTypeData(ExprType.number, Some(Number())))
-      case otherwise                => TypeInfo(expr, StaticTypeData(ExprType.string, None))
+      case IsNumberConstant(d) =>
+        TypeInfo(expr, StaticTypeData(ExprType.number, Some(Number(maxFractionalDigits = d.scale))))
```
josef-vlach, line 150: This seems like a sane thing to do and actually is not wrong, but due to the lack of any deeper expression specification wrt literal numbers, changing this will have potential effect across all the forms. So my question would be what is the motivation for this change?

It doesn't seems to be connected with GFORMS-1097 at all.

josef-vlach (changes requested):

PR #1094, opened 2021-03-05
GFORMS-1116 infoMessage in addToList summary

girishkamat:


app/uk/gov/hmrc/gform/views/form/addToList.scala.html
```diff
@@ -104,6 +105,11 @@
           </li>
 	}
       </ul>
+        @for(infoMessage <- bracket.source.infoMessage) {
+            <p>
```
josef-vlach, line 109: What is this `<p>` purpose?

 It can cause invalid html being rendered assuming that `@markDownParser(infoMessage)` will render some `div` as part of its content.

girishkamat: Didnt think too much, assumed <p> is a relative safe wrapper. Removed it

josef-vlach (approved):

PR #1101, opened 2021-03-10
GFORMS-804 - Fix issue of session ending once a user logs in to GG

SandipKhedekar:


app/uk/gov/hmrc/gform/playcomponents/GformSessionCookieCryptoFilter.scala
```diff
@@ -0,0 +1,90 @@
...
+
+    lazy val authconfigCookieValue = Crypted(rh.cookies.get(authConfigCookieName).fold("")(_.value))
+
+    val isPathParameterFormTemplateIdExists = rh.handlerDef.exists { hd =>
+      hd.parameterTypes.exists { _.getName() === FormTemplateIdClass }
+    }
+
+    val maybeFormTemplate: Future[Either[Unit, FormTemplate]] =
+      if (isPathParameterFormTemplateIdExists) {
+        val templateId = rh.uri.split("/")(3)
```
josef-vlach, line 62: `FormTemplateId` is not guaranteed to be on index `3`. We need to calculate index from `handlerDef` data.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/playcomponents/GformSessionCookieCryptoFilter.scala
```diff
@@ -0,0 +1,93 @@
...
+import uk.gov.hmrc.gform.sharedmodel.formtemplate.{ Anonymous, FormTemplate, FormTemplateId }
+import uk.gov.hmrc.http.HeaderCarrier
+import uk.gov.hmrc.play.HeaderCarrierConverter
+import uk.gov.hmrc.play.bootstrap.frontend.filters.crypto.{ SessionCookieCrypto, SessionCookieCryptoFilter }
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+class GformSessionCookieCryptoFilter(
+  sessionCookieCrypto: SessionCookieCrypto,
+  nonanonymouscookieCryptoFilter: SessionCookieCryptoFilter,
```
josef-vlach, line 37: Why is cookie here in lower case, whereas elsewhere it is typed with upper case?

Also rather than calling it `nonanonymouscookieCryptoFilter`  should better be to call it:
```
mdtpCookieCryptoFilter
```
or 
```
hmrcCookieCryptoFilter
```

app/uk/gov/hmrc/gform/playcomponents/GformSessionCookieCryptoFilter.scala
```diff
@@ -0,0 +1,93 @@
...
+import uk.gov.hmrc.gform.controllers.CookieNames._
+import uk.gov.hmrc.gform.gformbackend.GformConnector
+import uk.gov.hmrc.gform.sharedmodel.formtemplate.{ Anonymous, FormTemplate, FormTemplateId }
+import uk.gov.hmrc.http.HeaderCarrier
+import uk.gov.hmrc.play.HeaderCarrierConverter
+import uk.gov.hmrc.play.bootstrap.frontend.filters.crypto.{ SessionCookieCrypto, SessionCookieCryptoFilter }
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+class GformSessionCookieCryptoFilter(
```
josef-vlach, line 35: What about saying what this class is doing instead, something like:
```
SessionCookieDispatcherFilter
```
?

app/uk/gov/hmrc/gform/playcomponents/GformSessionCookieCryptoFilter.scala
```diff
@@ -0,0 +1,93 @@
...
+  sessionCookieCrypto: SessionCookieCrypto,
+  nonanonymouscookieCryptoFilter: SessionCookieCryptoFilter,
+  anonymousCookieCryptoFilter: SessionCookieCryptoFilter,
+  val sessionBaker: SessionCookieBaker,
+  gformConnector: GformConnector,
+  config: Config
+)(implicit override val ec: ExecutionContext, override val mat: Materializer)
+    extends SessionCookieCryptoFilter {
```
josef-vlach, line 43: This filter shouldn't handle session cookie itself, why it extends `SessionCookieCryptoFilter`?

app/uk/gov/hmrc/gform/playcomponents/GformSessionCookieCryptoFilter.scala
SandipKhedekar: Changed to **hmrcCookieCryptoFilter**. In fact, rename all nonanonymous*** names to hmrc***

app/uk/gov/hmrc/gform/playcomponents/GformSessionCookieCryptoFilter.scala
SandipKhedekar: Good. Changed.

app/uk/gov/hmrc/gform/playcomponents/GformSessionCookieCryptoFilter.scala
SandipKhedekar: At https://github.com/hmrc/gform-frontend/blob/master/app/uk/gov/hmrc/gform/playcomponents/FrontendFiltersModule.scala#L94  In bootstrap-frontend library class FrontendFilters 9th parameter as SessionCookieCryptoFilter hence extended SessionCookieCryptoFilter.

josef-vlach: My suggestion would be use original `SessionCookieCryptoFilter` in:
```
lazy val httpFilters: Seq[EssentialFilter] = new FrontendFilters(...) 
```
and then replace it with our `Filter`.  Something like:

```
httpFilters.map { 
  case _: SessionCookieCryptoFilter => ourFilter 
  case keep => keep
}
```

josef-vlach (approved):

PR #1114, opened 2021-03-15
GFORMS-804 - Fix issue of session ending once a user logs in to GG

SandipKhedekar:


josef-vlach:
So what is different, in this commit?

SandipKhedekar:
@josef-vlach I am decoding a Cookie at https://github.com/hmrc/gform-frontend/blob/GFORMS-804/app/uk/gov/hmrc/gform/playcomponents/SessionCookieDispatcherFilter.scala#L89 before decrypting it. There was an error **Unable to decrypt value** because of encoding.

josef-vlach (approved):

PR #1127, opened 2021-03-18
Suppress Cookie failed message authentication check logs in kibana

SandipKhedekar:


girishkamat (approved):

josef-vlach (changes requested):
This is not a fix of the problem

PR #1128, opened 2021-03-19
GFORMS-1153 - Improve performance by limiting data posted to the data…

josef-vlach:
… from the page

Support for "submitMode": "summaryinfoonly" date fields.

### DIFF
--- a/test/uk/gov/hmrc/gform/validation/DateValidationSpec.scala
+++ b/test/uk/gov/hmrc/gform/validation/DateValidationSpec.scala
@@ -16,588 +16,342 @@
 
 package uk.gov.hmrc.gform.validation
 
-/* import cats.instances.future._
- * import cats.scalatest.ValidatedValues._
- * import java.time.LocalDate
- * import org.scalatest.mockito.MockitoSugar.mock
- * import play.api.i18n.Messages
- * import uk.gov.hmrc.gform.Helpers.toSmartString
- * import uk.gov.hmrc.gform.auth.models.MaterialisedRetrievals
- * import uk.gov.hmrc.gform.fileupload.Envelope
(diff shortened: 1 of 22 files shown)


test/uk/gov/hmrc/gform/validation/DateValidationSpec.scala
```diff
@@ -16,588 +16,342 @@
...
+
+  implicit val l: LangADT = LangADT.En
+
+  private val lookupRegistry = new LookupRegistry(Map.empty)
+
+  implicit val smartStringEvaluator: SmartStringEvaluator = new SmartStringEvaluator {
+    override def apply(s: SmartString, markDown: Boolean): String = s.rawValue(LangADT.En)
+  }
+
+  implicit val messages: Messages = Helpers.stubMessages(
```
girishkamat, line 52: Minor: Could this be moved to a ExampleMessages trait, with all messages for test. I see this being duplicated in DateValidationLogicSpec

josef-vlach: That would tight couple those tests. Is that really what we want? In this way every test defined only what it needs, and it even clearly gives a hint what sort of keys are used by a particular test.

girishkamat: I know, I had that thought in mind. The messages are just constants and in all cases can represent real messages from the production file. The values themselves have no significance for tests no?

josef-vlach: > The messages are just constants and in all cases can represent real messages from the production file

Yes.

Values have significance for test, but they can be anything, they don't need to match real message.

girishkamat (approved):

PR #1143, opened 2021-03-31
GFORMS-1161 - Accessibility Issue focus on offscreen button fix

SandipKhedekar:


app/uk/gov/hmrc/gform/views/form/form.scala.html
```diff
@@ -69,11 +70,18 @@
 @formWithCSRF = @{new formWithCSRF()}
 @govukButton = @{new govukButton()}
 
-@displaySaveContinueButton(classes: String = "") = {
-    @if(shouldDisplayContinue) {
-        @govukButton(Button(content = Text(buttonLabel),
+@displaySaveContinueButton(classes: String = "") = @{
+    val attributes =
+      if(classes === "defaultsink") {
```
josef-vlach, line 75: Wouldn't clearer solution be to pass special attribute as an input?

Something like:
```
displaySaveContinueButton(classes: String = "", otherAttributes: Map[String, String] = Map.empty)  ...
```

girishkamat (approved):

PR #1146, opened 2021-04-01
Gforms 1175 fix duplicate file delete

girishkamat:


josef-vlach (approved):
Nothing wrong with doing this, but it is not a solution for GFORMS-1175.

As a matter of fact, I'd going to eliminate that `No associated FileId found` error altogether as a part of GFORMS-1055.

girishkamat:
I could reproduce this error quite easily on my machine, some times I clicked thrice and 3 errors were shown on the page. I think many users are used to the habit of double click on link and we were not preventing that from happening

josef-vlach:
As I said, that error is going to be eliminated once GFORMS-1055 is finished. But this still has a value though.

PR #1151, opened 2021-04-07
GFORMS-1182 - The focus is not set on the correct Date element on cli…

SandipKhedekar:
…cking the error message link

### DIFF
--- a/app/uk/gov/hmrc/gform/validation/ComponentsValidator.scala
+++ b/app/uk/gov/hmrc/gform/validation/ComponentsValidator.scala
@@ -116,7 +116,15 @@ class ComponentsValidator[D <: DataOrigin, F[_]: Monad](
   private def produceValidationError(
     message: SmartString
   ): ValidatedType[Unit] =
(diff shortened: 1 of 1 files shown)


app/uk/gov/hmrc/gform/validation/ComponentsValidator.scala
```diff
@@ -116,7 +116,15 @@ class ComponentsValidator[D <: DataOrigin, F[_]: Monad](
   private def produceValidationError(
     message: SmartString
   ): ValidatedType[Unit] =
-    Map(formComponent.modelComponentId -> Set(message.value)).invalid
+    formComponent.`type` match {
```
girishkamat, line 119: Can we not do it for any multivalue component? I have just implemented calendarDate and it also needs a similar change

SandipKhedekar: Done. Please check

girishkamat (approved):

PR #1153, opened 2021-04-07
GFORMS-1055 - Make the file upload component work so it does not use …

josef-vlach:
…javascript

### DIFF
--- a/app/uk/gov/hmrc/gform/sharedmodel/form/FormComponentIdToFileIdMapping.scala
+++ b/app/uk/gov/hmrc/gform/sharedmodel/form/FormComponentIdToFileIdMapping.scala
@@ -16,14 +16,13 @@
 
 package uk.gov.hmrc.gform.sharedmodel.form
 
-import cats.syntax.eq._
 import play.api.libs.json.Format
 import uk.gov.hmrc.gform.models.ids.ModelComponentId
 import uk.gov.hmrc.gform.sharedmodel.formtemplate.{ FormComponentId, JsonUtils }
 
 /** Keeps maping between FileUpload component and it's associated FileId in file-upload microservice.
   *
-  * When removing add-to-list iterations we
+  * When removing add-to-list iterations (or groups) we
   * are reindexing FormComponentIds of the components accross all iteration so there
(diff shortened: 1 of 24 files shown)


girishkamat (commented):
We get the following error when no file is chosen. Can this be improved?

`Only one file at a time can be uploaded`

josef-vlach:
In what way? Please read Slack discussion about this for some background.

app/uk/gov/hmrc/gform/sharedmodel/form/FormComponentIdToFileIdMapping.scala
```diff
@@ -39,19 +38,17 @@ case class FormComponentIdToFileIdMapping(mapping: Map[FormComponentId, FileId])
...
+      val inverse: Option[FileId] = inverseMapping.get(fileId).map(_.value).map(FileId.apply)
+      inverse.map { fileId =>
+        loop(fileId).getOrElse(fileId)
+      }
     }
-    FileId(mapping.get(formComponentId).fold(nextAvailable)(_.value))
 
+    FileId(
+      mapping.get(formComponentId).orElse(loop(FileId(formComponentId.value))).fold(formComponentId.value)(_.value)
+    )
```
girishkamat, line 51: Minor: the expression could be simplified

` mapping
      .getOrElse(formComponentId, 
          loop(FileId(formComponentId.value))
            .getOrElse(FileId(formComponentId.value)))`

public/javascripts/gformFileUpload.js
```diff
@@ -259,11 +259,14 @@
           interpolate(strings.maxSizeError[lang], [maxFileSize])
         );
       }
-
-      updateMapping($input.attr("id"), fileId, formTemplateId, accessCode)
+      uploadFile(file, fileId)
         .then(function(response) {
```
girishkamat, line 263: why was the order reversed?

girishkamat (approved):

public/javascripts/gformFileUpload.js
josef-vlach: To match order of events in no-js flow.

PR #1154, opened 2021-04-08
GFORMS-1177 ReferrerConfig exitMessage localized. 

girishkamat:
1. ReferrerConfig exitMessage localized
2. Clear auth entries from session after displaying acknowledgement page

### DIFF
--- a/app/uk/gov/hmrc/gform/controllers/GformSessionSupport.scala
+++ b/app/uk/gov/hmrc/gform/controllers/GformSessionSupport.scala
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2021 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.gform.controllers
+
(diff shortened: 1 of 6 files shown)


app/uk/gov/hmrc/gform/controllers/GformSessionSupport.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import play.api.mvc.{ AnyContent, Request, Result }
+import uk.gov.hmrc.gform.controllers.GformSessionKeys.{ EMAIL_AUTH_DETAILS_SESSION_KEY, REFERRER_CHECK_DETAILS }
+
+trait GformSessionSupport {
+  implicit class ResultOps(result: Result)(implicit request: Request[AnyContent]) {
+    def clearAuthFromSession() =
+      result.removingFromSession(EMAIL_AUTH_DETAILS_SESSION_KEY).removingFromSession(REFERRER_CHECK_DETAILS)
+  }
+}
```
josef-vlach, line 27: What about following cats ecosystem way of doing these?

```suggestion
trait GformSessionSyntax {
  implicit final def syntaxResultOps(result: Result): ResultOps = new ResultOps(result)
}

final class ResultOps(private val result: Result) extends AnyVal {
  def clearAuthFromSession(implicit request: Request[AnyContent]) =
    result.removingFromSession(EMAIL_AUTH_DETAILS_SESSION_KEY).removingFromSession(REFERRER_CHECK_DETAILS)
}
```

girishkamat: Not sure of the advantage of using it. Will add it anyways as it looks like a standard method

josef-vlach: The difference is that with `AnyVal` (cats) approach `clearAuthFromSession` will become static method, so `ResultOps` won't be instantiated when you want to call it.

 Whereas in original solution new instance of `ResultOps ` will be created anytime you are calling `clearAuthFromSession`.

josef-vlach (approved):

PR #1168, opened 2021-04-19
Gforms 1207 repeating page reduce count

girishkamat:


josef-vlach (changes requested):
We don't want to cleanse any user provided data if possible.

Problem is that we wrongly detect what data are hidden. So we should fix hidden fields detection instead, rather than cleansing data.

PR #1183, opened 2021-04-30
GFORMS-1196 - Include if on Acknowledgement page/section

SandipKhedekar:


josef-vlach (changes requested):
Comparison of `OptionResult` and `NumberResult` should always yield `false` as it doesn't make any sense.

There must be other fix for GFORMS-1196.

josef-vlach:
@SandipKhedekar I think fix would have to be done on backend. Look at [Rewriter.scala](https://github.com/hmrc/gform/blob/master/app/uk/gov/hmrc/gform/formtemplate/Rewriter.scala). 

For usability purposes we allow people to use `=` operator with a Choice component. For instance `areYouLivingInUK = 0`, but what they should do instead is to use `areYouLivingInUK contains 0` operator.

[Rewriter.scala](https://github.com/hmrc/gform/blob/master/app/uk/gov/hmrc/gform/formtemplate/Rewriter.scala) is doing rewrite of `=` to `contains` (where needed) so people don't need to do it. And I think this rewrite is missing for acknowledgement page.

PR #1186, opened 2021-05-05
GFORMS-1228 Handle choice elements, when they are mandatory but not selected by user in the form

girishkamat:


app/uk/gov/hmrc/gform/controllers/helpers/FormDataHelpers.scala
```diff
@@ -138,14 +146,39 @@ object FormDataHelpers {
...
+   * respective fields from the persistent form data
+   */
+  private def missingChoiceFieldIds(
+    maybeSectionNumber: Option[SectionNumber],
+    formModel: FormModel[DataExpanded],
+    requestData: Map[String, Seq[String]]
+  ) =
+    maybeSectionNumber.fold(Seq.empty[ModelComponentId]) { sectionNumber =>
+      formModel(sectionNumber).allFormComponents.collect {
+        case f @ IsChoice(_) if f.mandatory && !requestData.contains(f.id.value) =>
```
josef-vlach, line 176: Why `f. mandatory`? It should apply equally to optional choices.

app/uk/gov/hmrc/gform/controllers/helpers/FormDataHelpers.scala
```diff
@@ -138,14 +146,39 @@ object FormDataHelpers {
...
+        case ((variadicFormDataAcc, requestRelatedDataAcc), (maybeVar, maybeReq)) =>
+          (
+            maybeVar.fold(variadicFormDataAcc)(variadicFormDataAcc.addValue),
+            maybeReq.fold(requestRelatedDataAcc)(requestRelatedDataAcc + _)
+          )
+      }
+
+    (
+      variadicFormData,
+      missingChoiceFieldIds(maybeSectionNumber, formModel, requestData),
```
josef-vlach, line 160: I'm afraid that cleaning data here is not enough. They still live in `FormModelOptics`.

Problem can easily be show with this template:

```
{
  "title": "Multi-Value Revealing Choice",
  "fields": [
    {
      "id": "revealing",
      "type": "revealingChoice",
      "multivalue": "true",
      "label": "Revealing Choice",
      "mandatory": "no",
      "choices": ["Address"],
      "revealingFields": [
        [
          {
            "id": "fieldA",
            "type": "text",
            "label": "Field A",
            "format": "shortText"
          }
        ]
      ]
    }
  ]
},
{
  "title": "Page B",
  "includeIf": "${fieldA != ''}",
  "fields": [
    {
      "id": "fieldB",
      "type": "text",
      "label": "Field B",
      "format": "sterling"
    }
  ]
},
{
  "title": "Page C ${fieldB + 1000}",
  "fields": [
    {
      "id": "fieldC",
      "type": "text",
      "label": "Field C",
      "format": "shortText"
    }
  ]
}
```

1. Check `revealing`
2. Type something in `fieldA`
3. Type 123 in `fieldB`
4. See label say `Page C 1123`.
5. Go back to first page and untick `revealing`
6. Label should say `Page C 1000`, but it still shows `Page C 1123`.

Problem is that `FormModelOptics` has been created before this data cleaning happened.

josef-vlach (changes requested):
Only reliable solution for this issue I can see is to put all choices back into hidden fields.

But please read code comment and feel free to find other solution.

app/uk/gov/hmrc/gform/controllers/helpers/FormDataHelpers.scala
girishkamat: Hmm, you are right. This will make FormModelOptics stale for that execution

josef-vlach (approved):

PR #1187, opened 2021-05-07
GFORMS-1251 - Deterministic email code generation for Email Auth testing

SandipKhedekar:


conf/application.conf
```diff
@@ -49,6 +49,8 @@ tracking-consent-frontend {
   gtm.container = "a"
 }
 
+emailAuthEnabled = true
```
girishkamat, line 52: I think this name is misleading. May be we can call this `emailAuth.generateTestCode=true/false`

app/uk/gov/hmrc/gform/gform/EmailAuthController.scala
```diff
@@ -62,6 +63,8 @@ class EmailAuthController(
 
   private val logger: Logger = LoggerFactory.getLogger(getClass)
 
+  private val emailConfirmationCode = "GEAD"
```
girishkamat, line 66: Probably better to default to ABCD instead, easy to remember

josef-vlach: And also use `ci` right here ie `ci"ABCD"` to avoid that interpolation gymnastics down below: `ci"$emailConfirmationCode"`.

app/uk/gov/hmrc/gform/config/ConfigModule.scala
```diff
@@ -91,7 +91,8 @@ class ConfigModule(val context: ApplicationLoader.Context, playBuiltInsModule: P
         url       <- playConfiguration.getOptional[String]("optimizely.url")
         projectId <- playConfiguration.getOptional[String]("optimizely.projectId")
       } yield s"$url$projectId.js",
-      trackingConsentSnippet = new hmrcTrackingConsentSnippet(new TrackingConsentConfig(playConfiguration))
+      trackingConsentSnippet = new hmrcTrackingConsentSnippet(new TrackingConsentConfig(playConfiguration)),
+      emailAuthEnabled = typesafeConfig.getBoolean("emailAuth.generateTestCode")
```
girishkamat, line 95: this variable can we renamed as well

app/uk/gov/hmrc/gform/gform/EmailAuthController.scala
```diff
@@ -329,6 +332,12 @@ class EmailAuthController(
             )
           )
           .map(_ => emailAndCode)
+
+      case (_: EmailAuthConfig, false) =>
```
girishkamat, line 336: emailAuthEnabled should be renamed generateTestCode. When generateTestCode=true, we should response with ABCD, so this condition needs to be reversed

conf/application.conf
```diff
@@ -49,6 +49,8 @@ tracking-consent-frontend {
   gtm.container = "a"
 }
 
+emailAuth.generateTestCode = true
```
girishkamat, line 52: generateTestCode should be false by default. Any application that needs it should override it to true in the environment conf files

app/uk/gov/hmrc/gform/gform/EmailAuthController.scala
```diff
@@ -313,22 +316,33 @@ class EmailAuthController(
...
-              emailAndCode.code,
-              emailVerifierService
-            )
+
+        if (isDefaultEmailId) {
+          Future.successful(
+            EmailAndCode(ci"${emailId.value.toString}", EmailConfirmationCode(emailConfirmationCode))
           )
-          .map(_ => emailAndCode)
+        } else {
```
josef-vlach, line 327: I'd would prefer somehow more subtle approach. We don't need to have any code hardcoded etc. We can provide special RND to `emailVerificationCode ` which will give us deterministic code.

I did [POC of this approach](https://github.com/hmrc/gform-frontend/commit/dc36d8bfea28ec28c7cc99de0f0b8d1cd6a253a0).

app/uk/gov/hmrc/gform/config/FrontendAppConfig.scala
```diff
@@ -39,7 +39,8 @@ case class FrontendAppConfig(
   routeToSwitchLanguage: String => play.api.mvc.Call,
   contactFormServiceIdentifier: String,
   optimizelyUrl: Option[String],
-  trackingConsentSnippet: hmrcTrackingConsentSnippet
+  trackingConsentSnippet: hmrcTrackingConsentSnippet,
+  emailAuthDefaultEmailIds: Option[Seq[String]]
```
josef-vlach, line 43: `Option[Seq[String]]` is really weird type. Why not to make it `Option[NonEmptyList[CIString]]` ?

josef-vlach (commented):
Please look at my POC linked in one of the comment. 

Idea behind my POC is that we should attempt to always send an email, only for selected email addresses the code would be deterministic.

josef-vlach (approved):

PR #1190, opened 2021-05-11
GFORMS-1236 Back link works for non JS users

girishkamat:


public/javascripts/gformFormActionHandlers.js
```diff
@@ -45,7 +45,6 @@
 
 	    $("#main-content")
           .parent()
-          .on('click', '#backButton', setFormActionAndSubmit())
```
josef-vlach, line 48: Why we are removing this? I guess this is breaking save-on-back functionality?

girishkamat: Thats the question, do we need save-on-back functionality? Is this something they expected or something we delivered on our own? I remember Richie saying Back is just back, and not save and go back. I am trying to keep one route for back that 

1. only navigates as it says
2. uses GET as links are supposed to
3. works regardless of JS

josef-vlach: This ticket is not about changing behaviour of back link (when js is enabled). If we want to do that, that needs to be other discussion with the team.

girishkamat: The old functionality has been restored. Thanks

josef-vlach (approved):

PR #1201, opened 2021-05-18
GFORMS-752 Upgraded play-frontend-govuk and play-frontend-hrmc to latest. Added localization for character count messages in textarea fields

girishkamat:


josef-vlach (commented):
😱  if this is still not supported out of the box (ie, it is bug in the library) we shouldn't fix it on our side. That's why we use libraries, so we can delegate the work to others 👈

girishkamat:
We can say its not supported by the library and close this PR. Ofcourse the play upgrade is nice thing to still have as it fixed the name issue in hrmcDialog

josef-vlach (approved):

PR #1208, opened 2021-05-27
GFORMS-752 welsh translation

girishkamat:


conf/messages.cy
```diff
@@ -111,7 +111,9 @@ generic.shortText.error.pattern=Gall {0} ond cynnwys llythrennau, rhifau, bylcha
 generic.longText.error.pattern=Ni all {0} gynnwys y cymeriadau {1}
 generic.referenceNumber.error.pattern=Gall {0} ond cynnwys rhifau
 generic.charactersRemaining=Mae gennych {0} o gymeriadau yn weddill
+generic.characterRemaining=You have {0} character remaining
```
josef-vlach, line 114: Cannot we copy&paste (or rather reverse engineered) welsh from this: https://github.com/hmrc/play-frontend-hmrc/blob/master/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcCharacterCount.scala.html
?

girishkamat: Thats a great idea, I was about to send it for translation. Now I dont have to

girishkamat: The messages seem to be different though, nothing compared to what GDS provides

"You can enter up to...."

josef-vlach: If are going to ultimately switch to `HmrcCharacterCount` then it should be ok. But I'll leave that to you to decide.

josef-vlach (approved):

PR #1232, opened 2021-07-01
GFORMS-1299 - AddToList styling of list elements

SandipKhedekar:


public/stylesheets/gform.css
```diff
@@ -419,3 +419,5 @@ div.gform-date-input__item {
 .govuk-hint > p {
     margin: 0px;
 }
+
+.hmrc-add-to-a-list__identifier p{margin:0;}
```
josef-vlach, line 423: Can we format this in standard way?

SandipKhedekar: Done.

josef-vlach (approved):

PR #1240, opened 2021-07-09
GFORMS-1360 - Allow Welsh Toggle for Notify Emails.

josef-vlach:


app/uk/gov/hmrc/gform/sharedmodel/email/LocalisedEmailTemplateId.scala
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.gform.sharedmodel.email
+
+import play.api.libs.json.{ JsError, JsObject, JsString, JsSuccess, Reads }
+import uk.gov.hmrc.gform.sharedmodel.{ EmailVerifierService, LangADT, LocalisedString }
+import uk.gov.hmrc.gform.sharedmodel.notifier.NotifierTemplateId
+
+case class LocalisedEmailTemplateId(emailTemplateId: String, maybeEmailTemplateIdCy: Option[String]) {
```
girishkamat, line 23: LocalisedEmailTemplateId seems to be used in `UploadableEmailDestination` which used when uploading template in gform backend. Do we need this class and all the dependencies that come with it?

josef-vlach: No, we don't need them, but it's rabbit hole when I tried to remove them. We should create a new ticket for that.

girishkamat (approved):

PR #1242, opened 2021-07-12
GFORMS-1331 - For external services failure, log level should be ERROR

SandipKhedekar:


app/uk/gov/hmrc/gform/controllers/ErrResponder.scala
```diff
@@ -62,23 +62,31 @@ class ErrResponder(
...
     availableLanguages: Option[AvailableLanguages] = None
   )(implicit
     request: RequestHeader,
     messages: Messages
   ): Future[Result] = {
-    logger.info(s"Trying to access forbidden resource: $message")
+    if (isUpstreamError)
+      logger.error(s"Trying to access forbidden resource: $message")
+    else
+      logger.info(s"Trying to access forbidden resource: $message")
```
josef-vlach, line 88: To reduce duplications, can we refactor this logging logic into a class, something like:

```
class private CleverLogger(isUpstreamError: Boolean) {
  def log(msg: String) = if (isUpstreamError)
    logger.error(msg)
  else
    logger.info(msg)
  }
  
object CleverLogger {
    // constructors
    val LocalLogger = new CleverLogger(false)
    val UpstreamLogger = new CleverLogger(true)
}
```

and to pass `CleverLogger` around instead of `isUpstreamError: Boolean = false`

SandipKhedekar: Thanks @josef-vlach 
I have made changes as per suggestions. I have used the default parameter because defs `forbidden` and `forbiddenWithReason` in `ErrResponder` have been called from `AuthenticatedRequestActions` other than `ErrorHandler`.

josef-vlach (approved):

PR #1248, opened 2021-07-15
GFORMS-1359 Support for documentTitle on all pages

girishkamat:


josef-vlach:
Rather than using `documentTitle` wouldn't better name be `noPiiTitle`?

Who would guess at first glance what difference between `title` and `documentTitle` is, or which is which?

I think more descriptive name would cause significantly less confusion going on.

girishkamat:
> Rather than using `documentTitle` wouldn't better name be `noPiiTitle`?
> 
> Who would guess at first glance what difference between `title` and `documentTitle` is, or which is which?
> 
> I think more descriptive name would cause significantly less confusion going on.

Done

josef-vlach (approved):

PR #1283, opened 2021-08-20
GFORMS-1417 Added support for referring repeating variables from outside

girishkamat:


app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -54,18 +54,35 @@ case class EvaluationResults(
...
   ): ExpressionResult = {
     val modelComponentId = expr.formComponentId.modelComponentId
-    val expressionResult = exprMap.getOrElse(
-      expr,
-      recData.variadicFormData
-        .get(modelComponentId)
-        .fold(ExpressionResult.empty)(fromVariadicValue)
+    val isModelFormComponentIdPure = modelComponentId.indexedComponentId.isPure
+    val isReferenceIndexed = evaluationContext.indexedComponentIds.exists(
+      _.baseComponentId == modelComponentId.baseComponentId
```
josef-vlach, line 62: `===`

app/uk/gov/hmrc/gform/eval/ExpressionResult.scala
```diff
@@ -123,6 +134,7 @@ sealed trait ExpressionResult extends Product with Serializable {
     case t: DateResult    => false
     case t: PeriodResult  => er.ifPeriodResult(t.value.toTotalMonths == _.toTotalMonths)
     case t: AddressResult => false
+    case t: ListResult    => false
```
josef-vlach, line 137: We should provide meaningful implementation for all operation.

This applies for all other 
```
case t: ListResult    => false
```

josef-vlach (changes requested):
Also when indexed field is printed in SmartString it prints out the answers in wrong order with no spaces between `,` that is if comma is not missing altogether (as in case of sterling).

app/uk/gov/hmrc/gform/eval/ExpressionResult.scala
girishkamat: Agreed, just added an implementation with tests

girishkamat:
> Also when indexed field is printed in SmartString it prints out the answers in wrong order with no spaces between `,` that is if comma is not missing altogether (as in case of sterling).

Sorted the values and printed with space between commas

app/uk/gov/hmrc/gform/eval/ExpressionResult.scala
```diff
@@ -101,10 +106,16 @@ sealed trait ExpressionResult extends Product with Serializable {
     case t: OptionResult =>
       er.fold[Boolean](_ => false)(_ => false)(_ => false)(n => t.contains(n.value))(_ => false)(_ => false)(_ =>
         false
-      )(_ => false)(_ => false)
+      )(_ => false)(_ => false)(_ => false)
     case t: DateResult    => false
     case t: PeriodResult  => false
     case t: AddressResult => false
+    case t: ListResult =>
```
josef-vlach, line 113: Can't this be simply 
```
t.list.exists(_.contains(er))
```
So the `exists` semantics is clearly visible?

app/uk/gov/hmrc/gform/eval/ExpressionResult.scala
```diff
@@ -123,6 +134,7 @@ sealed trait ExpressionResult extends Product with Serializable {
     case t: DateResult    => false
     case t: PeriodResult  => er.ifPeriodResult(t.value.toTotalMonths == _.toTotalMonths)
     case t: AddressResult => false
+    case t: ListResult    => er.ifListResult(_ === t.list)
```
josef-vlach, line 137: There seems to be no `exists` semantics present here.
Shouldn't implementation be:
```
t.list.exists(_.identical(er))
```

app/uk/gov/hmrc/gform/eval/ExpressionResult.scala
```diff
@@ -135,6 +147,7 @@ sealed trait ExpressionResult extends Product with Serializable {
     case t: DateResult    => false
     case t: PeriodResult  => er.ifPeriodResult(t.value.toTotalMonths > _.toTotalMonths)
     case t: AddressResult => false
+    case t: ListResult    => er.ifNumberResult(n => t.list.exists(_.ifNumberResult(_ > n)))
```
josef-vlach, line 150: This is missing `>` operation on String, Period and Option results.
Again better implementation seems to be
```
t.list.exists(_.>(er))
```

app/uk/gov/hmrc/gform/eval/ExpressionResult.scala
```diff
@@ -147,6 +160,7 @@ sealed trait ExpressionResult extends Product with Serializable {
     case t: DateResult    => false
     case t: PeriodResult  => er.ifPeriodResult(t.value.toTotalMonths >= _.toTotalMonths)
     case t: AddressResult => false
+    case t: ListResult    => er.ifNumberResult(n => t.list.exists(_.ifNumberResult(_ >= n)))
```
josef-vlach, line 163: Similar as cases above?

Applies for the rest of `case t: ListResult` cases

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/eval/ExpressionResult.scala
girishkamat: Not sure how useful this one is, I was just checking the list were having identical values i.e equality. Checking for contains when it says identical will be confusing may be?

app/uk/gov/hmrc/gform/eval/ExpressionResult.scala
girishkamat: Hmm, I didnt want it for other values as > is mainly used for numbers only. Not sure why one would use > for string or option, but Period will be useful. Your impl seems to handle all

app/uk/gov/hmrc/gform/eval/ExpressionResult.scala
girishkamat: Thanks, will do

app/uk/gov/hmrc/gform/eval/ExpressionResult.scala
girishkamat: This doesnt seem to work, as contains for StringResult, NumberResult etc seems to default to false

josef-vlach: `contains` operator is used only in conjunction with choice/revealingChoice components, so no `StringResult` or `NumberResult` should be considered and thus false is correct behaviour.
IOW `${surname contains 'Smith'}` is invalid expression if surname is `type: text`.

app/uk/gov/hmrc/gform/eval/ExpressionResult.scala
josef-vlach: With this implementation only expressions of this form can ever be true:
```
${amount = amount}
```
This looks very limiting

More reasonable expressions which should potentially evaluate to true are of the form:
```
${amount = 1000}
```

app/uk/gov/hmrc/gform/eval/ExpressionResult.scala
girishkamat: Yes, for that reason when 

```t.list.exists(_.contains(er))```

evaluates and runs contains on StringResult, it returns false and fails.

File: app/uk/gov/hmrc/gform/eval/ExpressionResult.scala (hunk 4 above)
Comment (josef-vlach) [reply app/uk/gov/hmrc/gform/eval/ExpressionResult.scala:137]: Perhaps best implementation is to support both cases:

```
er.ifListResult(_ === t.list) || t.list.exists(_ identical er)
```

girishkamat: Ahh, now it makes sense. Done

josef-vlach (approved):

PR #1291, opened 2021-09-03
GFORMS-1377 - Composite Auth

SandipKhedekar:


app/uk/gov/hmrc/gform/views/hardcoded/CompositeAuthFormPage.scala
```diff
@@ -0,0 +1,118 @@
...
+                value = Some(id),
+                content = Text(messages("compositeAuth.ggContinueContent"))
+              ),
+              RadioItem(
+                value = Some(configName),
+                content = Text(messages("compositeAuth.ggDifferentContent")),
+                conditionalHtml = Some(
+                  pWithStrong(
+                    messages("compositeAuth.ggDifferentConditional"),
+                    id.replaceAll("..(?!$)", "$0 "),
```
josef-vlach, line 84: This is used on two places, it would be good to extract this `replaceAll` as utility method and to perhaps write test for it, as it is not obvious what this is supposed to achieve.

josef-vlach (changes requested):
Minor details + there is Welsch translation missing.

app/uk/gov/hmrc/gform/gform/CompositeAuthController.scala
```diff
@@ -0,0 +1,103 @@
...
+        .fold(
+          errorForm => {
+            val compositeAuthFormPage = new CompositeAuthFormPage(formTemplate, errorForm, None)
+            BadRequest(html.auth.auth_selection(frontendAppConfig, compositeAuthFormPage, None)).pure[Future]
+          },
+          {
+            case AuthConfig.hmrcSimpleModule =>
+              val formTemplate = request.attrs(FormTemplateKey)
+              val continueUrl =
+                s"${frontendAppConfig.gformFrontendBaseUrl}${uk.gov.hmrc.gform.gform.routes.NewFormController.dashboardWithCompositeAuth(formTemplate._id).url}"
```
josef-vlach, line 85: Wouldn't be simpler to use `+` instead of string interpolation?
```
 frontendAppConfig.gformFrontendBaseUrl + uk.gov.hmrc.gform.gform.routes.NewFormController.dashboardWithCompositeAuth(formTemplate._id).url
 ```

josef-vlach (changes requested):

josef-vlach (approved):

PR #1294, opened 2021-09-07
GFORMS-821 Fixed issue with ${link.newForm.xxx}

girishkamat:


josef-vlach:
Why is commit message
```
GFORMS-1252 Email Keeping Users Logged In even after form submission
```
?

girishkamat:
Just updated with the right commit message, thanks

josef-vlach (approved):

PR #1295, opened 2021-09-08
Gforms 1457 default researchbanner beta

girishkamat:


SandipKhedekar:
@girishkamat I believe, we need to keep the code for ResearchBanner so we can use it when we want to render it specifically. 
Acceptance Criteria is not clear hence asking.

SandipKhedekar (approved):

PR #1298, opened 2021-09-09
GFORMS-1252 Email Keeping Users Logged In even after form submission …

girishkamat:
…- bug fix


app/uk/gov/hmrc/gform/gform/NewFormController.scala
```diff
@@ -18,7 +18,6 @@ package uk.gov.hmrc.gform.gform
```
girishkamat, line 18: reverting this class to the previous state. All logic for clearing session is moved to EmailAuthSessionPurgeFilter.scala

josef-vlach (approved):

PR #1302, opened 2021-09-13
GFORMS-1466 Implemented expression user.enrolments.[SERVICE].[IDENTIFIER].count/index

girishkamat:


app/uk/gov/hmrc/gform/graph/processor/UserCtxEvaluatorProcessor.scala
```diff
@@ -32,8 +32,25 @@ object UserCtxEvaluatorProcessor extends IdentifierExtractor {
     (retrievals, userField) match {
       case (AuthenticatedRetrievals(_, enrolments, _, _, _), UserField.EnrolledIdentifier) =>
         authorizedEnrolmentValue(enrolments, authConfig)
-      case (AuthenticatedRetrievals(_, enrolments, _, _, _), UserField.Enrolment(sn, in)) =>
-        extractIdentifier(enrolments, sn, in)
```
josef-vlach, line 36: No one is using `extractIdentifier` now, so it should be removed

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Expr.scala
```diff
@@ -202,11 +210,19 @@ object RosmProp {
   implicit val format: OFormat[RosmProp] = derived.oformat()
 }
 
-sealed trait UserField
+sealed trait UserField {
+  def fold[B](f: => B)(g: UserField.Enrolment => B)(h: => B): B =
```
josef-vlach, line 214: From caller point of view this seems as unsafe definition. There is nothing to say what `f` vs `h` cases stands for.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Expr.scala
girishkamat: valid point, so it should be something like

```
def fold[B](f: UserField.AffinityGroup.type => B)(g: UserField.Enrolment => B)(h: UserField.Something.type => B): B =
```

girishkamat: Done

app/uk/gov/hmrc/gform/graph/processor/UserCtxEvaluatorProcessor.scala
girishkamat: Removed, thanks

josef-vlach (approved):

PR #1305, opened 2021-09-15
GFORMS-1463 - Extend validation in Add to List to allow sum reference

josef-vlach:


app/uk/gov/hmrc/gform/sharedmodel/graph/DependencyGraph.scala
```diff
@@ -72,12 +73,17 @@ object DependencyGraph {
         case otherwise                => List.empty
       }
 
+    def eqBaseComponentId(expr: Expr, fc: FormComponent): Boolean = expr
+      .cast[FormCtx]
+      .map(_.formComponentId.baseComponentId === fc.id.baseComponentId)
+      .contains(true)
```
girishkamat, line 79: cant this be 

```
expr.cast[FormCtx].exists(_.formComponentId.baseComponentId === fc.id.baseComponentId)
```

test/uk/gov/hmrc/gform/graph/DependencyGraphSuite.scala
```diff
@@ -0,0 +1,176 @@
+/*
```
girishkamat, line 1: we have a DependencyGraphSpec class already. If we plan to use msuite, we need to migrate all tests to the new structure some day

app/uk/gov/hmrc/gform/sharedmodel/graph/DependencyGraph.scala
josef-vlach: Good catch

test/uk/gov/hmrc/gform/graph/DependencyGraphSuite.scala
josef-vlach: > we have a DependencyGraphSpec class already

Yes I know, unfortunately the nature of it isn't exactly what I need and rather than bloat it decided ti create independent Suite.
We may let scala-test and munit to live side by side imho.

girishkamat (approved):

PR #1306, opened 2021-09-17
GFORMS-1455 - DAC: Implement an in progress & completed status for fi…

josef-vlach:
…le upload

### DIFF
--- a/public/javascripts/gformFileUpload.js
+++ b/public/javascripts/gformFileUpload.js
@@ -317,7 +317,7 @@
       );
       $("#" + formComponentId + "-files")
         .empty()
-        .append(startProgressBar(""));
+        .append(startProgressBar());
       return $.ajax({
         url:
           "/file-upload/upload/envelopes/" +
--- a/public/javascripts/gformFileUpload.js
+++ b/public/javascripts/gformFileUpload.js
@@ -28,6 +28,14 @@
     deleteLabel: {
       en: "Delete",
       cy: "Dileu"
+    },
+    uploadingFile: {
+      en: "Uploading file",
+      cy: "Uploading file"
+    },
+    hasBeenUploaded: {
+      en: "has been uploaded",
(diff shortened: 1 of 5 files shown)


public/javascripts/gformFileUpload.js
```diff
@@ -28,6 +28,14 @@
...
       en: "Delete",
       cy: "Dileu"
+    },
+    uploadingFile: {
+      en: "Uploading file",
+      cy: "Uploading file"
+    },
+    hasBeenUploaded: {
+      en: "has been uploaded",
+      cy: "has been uploaded"
```
girishkamat, line 38: I know we dont have to solve it now, but it would be good to initialize these labels from the messages properties file on page load, so we dont store translations in 2 different places (like we did in https://github.com/hmrc/gform-frontend/blob/master/app/uk/gov/hmrc/gform/views/govuk_wrapper.scala.html, but allow each page to define what it needs to load)

public/javascripts/gformFileUpload.js
```diff
@@ -298,13 +308,16 @@
...
+    function uploadFile(file, fileId, formComponentId) {
       var formData = new FormData();
       formData.append(
         fileId,
         file,
         fileId + "_" + file.name.replace(/\\/g, "/").replace(/.*\//, "")
       );
+      $("#" + formComponentId + "-files")
+        .empty()
+        .append(startProgressBar(""));
```
girishkamat, line 320: startProgressBar is a no-arg function, why passing ""?

josef-vlach: Fixed

public/javascripts/gformFileUpload.js
josef-vlach: Good call, I've made a note to fix this in upscan related work

girishkamat (approved):

PR #1311, opened 2021-09-21
GFORMS-1377 - Error page Welsh translation fix

SandipKhedekar:


app/uk/gov/hmrc/gform/gform/CompositeAuthController.scala
```diff
@@ -50,11 +51,21 @@ class CompositeAuthController(
...
+    ggId: Option[String],
+    continue: String,
+    se: SuppressErrors
+  ) =
     nonAutheticatedRequestActions.async { implicit request => implicit lang =>
       val formTemplate = request.attrs(FormTemplateKey)
       val compositeAuthFormPage =
-        new CompositeAuthFormPage(formTemplate, choice, ggId)
+        choice
+          .bindFromRequest()
```
josef-vlach, line 64: Is this binding needed? Will it ever bind successfully given this is a GET endpoint?

app/uk/gov/hmrc/gform/views/hardcoded/CompositeAuthFormPage.scala
```diff
@@ -51,7 +53,10 @@ class CompositeAuthFormPage(
     )
   }
 
-  val hasErrors: Boolean = errorSummary.errorList.nonEmpty
+  val hasErrors: Boolean =
+    if (se.asString === SuppressErrors.seNo)
```
josef-vlach, line 57: Descending to `String` level is quite ugly, what about implementing `Eq[SupressError]` typeclass instance instead?

josef-vlach (changes requested):

josef-vlach (approved):

PR #1323, opened 2021-10-11
GFORMS-1518 - Do not show phase banner for live forms.

rhubner:


DursunGundogan (approved):

app/uk/gov/hmrc/gform/views/govuk_wrapper.scala.html
```diff
@@ -143,13 +143,13 @@
 }
 
 @phaseBanner = @{
-  developmentPhase.fold(SiteBanner(Some(BetaBanner.banner))){ dp =>
+  developmentPhase.fold[Option[PhaseBanner]](Some(SiteBanner(Some(BetaBanner.banner), HtmlContent(feedbackContent)))){ dp =>
```
josef-vlach, line 146: Can we  construct beta banner only once and reuse it on both places, so that people do not forget to modify one or the other in the future?

rhubner: No, _Alpha/Beta_ case is using `banner` property from `DevelopmentPhase` trait. And this in not available in default section in `fold`.

josef-vlach: All I'm trying to say is why not to make `PhaseBanner` a function of `DevelopmentPhase`?

ie. `DevelopmentPhase => Option[PhaseBanner]`

Implementation:
```
import uk.gov.hmrc.govukfrontend.views.viewmodels.content.HtmlContent
import uk.gov.hmrc.govukfrontend.views.viewmodels.phasebanner.PhaseBanner
import uk.gov.hmrc.gform.views.components.SiteBanner

sealed trait DevelopmentPhase {
  val banner: String

  def toPhaseBanner(feedbackContent: Html): Option[PhaseBanner] = this match {
    case AlphaBanner    => Some(SiteBanner(Some(banner), HtmlContent(feedbackContent)))
    case BetaBanner     => Some(SiteBanner(Some(banner), HtmlContent(feedbackContent)))
    case ResearchBanner => Some(SiteBanner(Some(banner)))
    case LiveBanner     => None
  }
}
```
And use site will become only:
```
phaseBanner = developmentPhase.getOrElse(BetaBanner).toPhaseBanner(feedbackContent),
```

josef-vlach (approved):

PR #1326, opened 2021-10-12
GFORMS-1492 - Introduce a new property to store old formId(s) of form to retrieve saved data

DursunGundogan:


app/uk/gov/hmrc/gform/gformbackend/GformConnector.scala
```diff
@@ -119,12 +119,31 @@ class GformConnector(ws: WSHttp, baseUrl: String) {
...
+              def getFirstTemplateByLegacyIds(legacyIds: List[FormTemplateId]): Future[Option[FormTemplate]] =
+                legacyIds match {
+                  case Nil => Future.successful(None)
+                  case legacyFormId :: tail =>
+                    getFormTemplate(legacyFormId).map(Some(_)).recoverWith {
+                      case UpstreamErrorResponse.WithStatusCode(statusCode, _)
+                          if statusCode == StatusCodes.NotFound.intValue =>
+                        getFirstTemplateByLegacyIds(tail)
+                    }
+                }
```
josef-vlach, line 136: We shouldn't do any assumption about existence of `FormTemplate`. Or rather we should assume it does not exist, but user data can still be linked to the nonexistent `FormTemplate`. 

In other word need to check all `formTemplate.legacyFormIds` for existence of user data without making any assumption about corresponding `FormTemplate`.

app/uk/gov/hmrc/gform/gformbackend/GformConnector.scala
```diff
@@ -119,12 +119,31 @@ class GformConnector(ws: WSHttp, baseUrl: String) {
...
+
+              val legacyFormIds = formTemplate.legacyFormIds.map(_.toList).getOrElse(Nil)
+
+              OptionT(getFirstTemplateByLegacyIds(legacyFormIds)).flatMap { formTemplate =>
+                val updatedFormIdData = formIdData.withTemplateId(formTemplate)
+                OptionT(maybeForm(updatedFormIdData, formTemplate)).semiflatTap { form =>
+                  createFormFromLegacy(updatedFormIdData, formIdData)
+                  Future.successful(form.copy(_id = formIdData.toFormId, formTemplateId = formTemplate._id))
```
josef-vlach, line 144: This looks brittle. Parallel computation where "hard to reason about things" can happen. Let's make `createFormFromLegacy` to return `Form` which will be returned as a result.

app/uk/gov/hmrc/gform/gformbackend/GformConnector.scala
```diff
@@ -85,6 +85,38 @@ class GformConnector(ws: WSHttp, baseUrl: String) {
...
+                  }
+                }
+                .recoverWith {
+                  case UpstreamErrorResponse.WithStatusCode(statusCode, _)
+                      if statusCode == StatusCodes.NotFound.intValue =>
+                    getFormByLegacyIds(tail)
+                }
+          }
+
+        val legacyFormIds = formTemplate.legacyFormIds.map(_.toList).getOrElse(Nil)
```
josef-vlach, line 113: You've decided to make `legacyFormIds` of type `Option[NonEmptyList[FormTemplateId]]` which is great, but as soon as you can use that representation to your advantage you are throwing it away by converting it to `List[FormTemplateId]`?

Why not to use `legacyFormIds.fold(nothing-to-do-here)(do-the-logic-here)` pattern instead and then `getFormByLegacyIds` can become:

```
def getFormByLegacyIds(legacyIds: NonEmptyList[FormTemplateId]): Future[Option[Form]]
```

app/uk/gov/hmrc/gform/gformbackend/GformConnector.scala
```diff
@@ -85,6 +85,38 @@ class GformConnector(ws: WSHttp, baseUrl: String) {
...
     hc: HeaderCarrier,
     ec: ExecutionContext
+  ): Future[Option[Form]] =
+    findForm(formIdData, formTemplate).recoverWith {
+      case UpstreamErrorResponse.WithStatusCode(statusCode, _) if statusCode == StatusCodes.NotFound.intValue =>
+        def getFormByLegacyIds(legacyIds: List[FormTemplateId]): Future[Option[Form]] =
+          legacyIds match {
+            case Nil => Future.successful(None)
+            case legacyFormId :: tail =>
+              getFormTemplate(legacyFormId)
```
josef-vlach, line 95: I still find it weird we are doing this. I think we shouldn't assume old form template is still present in db.

Only reason why you need full `FormTemplate` is to pass it `findForm` method, but I think we should not use `findForm` method for this legacy ids and use directly `getForm` instead.

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/FormTemplate.scala
```diff
@@ -26,6 +26,7 @@ import uk.gov.hmrc.gform.sharedmodel.formtemplate.destinations.Destinations
 case class FormTemplate(
   _id: FormTemplateId,
   originalId: FormTemplateId,
+  legacyFormIds: Option[NonEmptyList[FormTemplateId]],
```
josef-vlach, line 29: Integration test are not passing at the moment, due to this change, please run `it:test` and fix compilation errors.

josef-vlach (changes requested):

PR #1334, opened 2021-10-18
GFORMS-1505 - Add splunk events for monitoring user time taken to complete a form

DursunGundogan:


app/uk/gov/hmrc/gform/auditing/AuditService.scala
```diff
@@ -41,21 +41,34 @@ trait AuditService {
       modelComponentId.toMongoIdentifier -> variadicValue.toSeq.mkString(",")
     }.toMap
 
+  def sendFormSaveEvent[D <: DataOrigin](
+    form: Form,
+    retrievals: MaterialisedRetrievals
+  )(implicit ec: ExecutionContext, hc: HeaderCarrier): Unit =
+    sendEvent("formSaved", form, Map.empty, retrievals, CustomerId.empty)
```
josef-vlach, line 48: Can we replace `save` with `started` to make it more clear that this is used only at the start of the form?

PR #1346, opened 2021-11-02
GFORMS-1549 - Support non-javascript upload to Upscan

DursunGundogan:


app/uk/gov/hmrc/gform/upscan/UpscanController.scala
```diff
@@ -120,9 +122,22 @@ class UpscanController(
...
-        val flash = mkFlash("file.error.upload.one.only")
         val fileId = FileId(formComponentId.value)
 
+        val flash = errorCode match {
+          case Some("InvalidArgument") =>
+            mkFlash(
+              "generic.error.upload",
+              formModelOptics.formModelVisibilityOptics.fcLookup
+                .get(formComponentId)
+                .map(_.label.localised.value)
```
josef-vlach, line 133: To get correct label (which applies all the expression etc.)  `.value` on `SmartString` needs to be used

```
.map(_.label.value)
```
To make this compile, ad this import 
```
import uk.gov.hmrc.gform.eval.smartstring.SmartStringEvaluationSyntax
```

and replace `_` by `implicit sse` in callbacks params

```
implicit request => implicit l => cache => implicit sse => implicit formModelOptics =>
```

josef-vlach (changes requested):

josef-vlach (approved):

PR #1352, opened 2021-11-08
GFORMS-1550 - Improve Upscan error message

rhubner:


public/javascripts/gformFileUpload.js
```diff
@@ -298,7 +302,9 @@
           }, onError)
           .then(function(response) {
             if(response == "error") {
```
josef-vlach, line 304: I think this PR should be postponed after work on non-js journey is completed.

I believe result of that work will be that `UpscanFileStatus.Failed` will carry exact reason for an error, is that right @DursunGundogan 

So instead of vague `response == "error"` we will be able to display correct error message, as opposite to always fallback to `strings.upscanError[lang]`

DursunGundogan: Right, `UpscanFileStatus.Failed` will carry the exact reason for an error after that work.

josef-vlach (approved):

PR #1357, opened 2021-11-10
GFORMS-1549 - Support non-javascript upload to Upscan

DursunGundogan:


app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -384,8 +384,22 @@ class SectionRenderingService(
...
+    val formModel = formModelOptics.formModelRenderPageOptics.formModel
+    val pageModel = formModel(sectionNumber).allFormComponents
+    val componentSize = pageModel.map(_.id).size
+
+    val fileUploadProviders: List[FileUploadProvider] = pageModel.collect { case IsFileUpload(fu) =>
+      fu.fileUploadProvider
+    }
+
+    val actionForm =
+      if (fileUploadProviders.size === 1 && componentSize === 1) {
```
josef-vlach, line 397: `componentSize === 1` is redundant

app/uk/gov/hmrc/gform/views/form/form.scala.html
```diff
@@ -124,7 +125,7 @@
       }
     }
 
-    @formWithoutCSRF(action = page.formAction, args = 'class -> "js-form", 'autocomplete -> "on", 'novalidate -> "novalidate", 'id -> "gf-form") {
+    @formWithCSRF(action = page.formAction, args = 'class -> "js-form", 'autocomplete -> "on", 'novalidate -> "novalidate", 'id -> "gf-form") {
```
josef-vlach, line 128: How come this is hardcoded to `formWithCSRF`? This will still send `csrfToken` to upscan which is going to fail. Or am I missing something?

Should this be managed by same condition as `formAction`:
```
if (fileUploadProviders.size === 1) formWithoutCSRF else formWithCSRF
```

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/views/form/form.scala.html
```diff
@@ -124,25 +126,48 @@
       }
     }
 
-    @formWithoutCSRF(action = page.formAction, args = 'class -> "js-form", 'autocomplete -> "on", 'novalidate -> "novalidate", 'id -> "gf-form") {
+    @if(page.renderWithoutCSRF){
```
josef-vlach, line 129: This condition is redundant. It is same as `page.upscanUrl.isDefined`.
When we have  `page.upscanUrl` we want to render `"gf-upscan"`

Also `"gf-form"` form should be defined only once in this file.

public/javascripts/gformFileUpload.js
```diff
@@ -343,7 +343,7 @@
 
     // Handle upscan file upload request
     function uploadFileUpscan(dataset) {
-      var formData = new FormData(document.getElementById("gf-form"));
+      var formData = new FormData(document.getElementById("gf-upscan"));
```
josef-vlach, line 346: Isn't `file-upload-frontend` journey affected by this change?

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -384,6 +384,18 @@ class SectionRenderingService(
       formLevelHeading,
       specialAttributes = Map.empty
     )
+
+    val formModel = formModelOptics.formModelRenderPageOptics.formModel
+    val pageModel = formModel(sectionNumber).allFormComponents
```
josef-vlach, line 389: This should be named `formComponents`

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -384,6 +384,18 @@ class SectionRenderingService(
...
+
+    val fileUploadProviders: List[FileUploadProvider] = pageModel.collect { case IsFileUpload(fu) =>
+      fu.fileUploadProvider
+    }
+
+    val upscanUrl: Option[String] = if (fileUploadProviders.size === 1) {
+      Some(upscanInitiate.get(pageModel.find(_.`type`.isInstanceOf[FileUpload]).get.id).uploadRequest.href)
```
josef-vlach, line 396: This will not work in Add To List, it will find wrong upload  request.

Another issue is that this will fail on `file-upload-frontend` journey. There is missing check on correct provider, ie `FileUploadProvider.Upscan`.

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -1096,6 +1110,7 @@ class SectionRenderingService(
       inputType = Some("submit"),
       classes = "govuk-button--secondary",
       attributes = Map(
+        "form"        -> "gf-upscan",
```
josef-vlach, line 1113: Isn't `file-upload-frontend` journey affected by this change?

josef-vlach (changes requested):
We need to make sure `file-upload-frontend` based journey is not broken.

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -766,6 +791,64 @@ class SectionRenderingService(
...
+      .map(_.fileId)
+      .getOrElse(ei.envelope.mapping.fileIdFor(formComponent.id))
+
+  private def isVisible(formComponent: FormComponent, ei: ExtraInfo): Boolean = formComponent.includeIf.fold(true) {
+    includeIf =>
+      ei.formModelOptics.formModelVisibilityOptics.evalIncludeIfExpr(includeIf, None)
+  }
+
+  private def htmlForUpscan(
+    renderUnit: RenderUnit,
```
josef-vlach, line 806: This should not use `renderUnit: RenderUnit`. It should use `formComponent: FormComponent` which we know is upscan "file"

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -766,6 +791,64 @@ class SectionRenderingService(
...
+            fileUploadProvider match {
+              case FileUploadProvider.Upscan =>
+                val upscanInitiateResponse: UpscanInitiateResponse = upscanInitiate.get(formComponent.id)
+                val hiddenFields: List[Html] = upscanInitiateResponse.uploadRequest.fields.toList.map {
+                  case (name, value) =>
+                    html.form.snippets.hidden(name, value)
+                }
+                val fileUploadName = "file"
+                val attributes = Map("upscan" -> "upscan")
+                htmlForFileUpload(
```
josef-vlach, line 832: I think we should render the final form right here ie. `formWithoutCSRF`. That form should be bare form with only `hiddenFields` and nothing else.

We should also capture metadata about this form. ie. `(fcId: FormComponentId, htmlId: String)`. We will need this once we render main form `gf-form`

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -833,32 +912,8 @@ class SectionRenderingService(
...
-                  formComponent,
-                  formTemplateId,
-                  ei,
-                  validationResult,
-                  fileId,
-                  fileUploadName,
-                  upscanInitiateResponse.uploadRequest.href,
-                  attributes,
-                  hiddenFields
-                )
```
josef-vlach, line 861: We should keep this as it was, we will only need to link it to correct `<form> with htmlId` from before.

josef-vlach (changes requested):
I was surprised you didn't use `form` attribute to link to upscan form at all, but than it dawn on me. This solution is not keeping order of the component as defined in the template, which is not ideal.
For example there may be some info component designers wants to render above file component, but this solution will revert that order. I think we should fix it. 

I've wrote down couple of notes how it should work. Main piece to change is `htmlForUpscan`, but as a result it should be drastically simpler and more focused.

app/uk/gov/hmrc/gform/views/form/form.scala.html
```diff
@@ -124,25 +126,32 @@
...
+    }
+    @formWithCSRF(action = page.formAction, args = 'class -> "js-form", 'autocomplete -> "on", 'novalidate -> "novalidate", 'id -> "gf-form") {
+        @{page.snippets}
+    }
 
-      <div class="govuk-button-group">
-          @displaySaveContinueButton()
-          @if(page.renderComeBackLater) {
+    <div class="govuk-button-group">
+        @displaySaveContinueButton(otherAttributes = Map("form" -> "gf-form"))
```
josef-vlach, line 143: Can we undo this change and move it back to `gf-form` form?

app/uk/gov/hmrc/gform/views/form/form.scala.html
```diff
@@ -124,25 +126,32 @@
...
+        @if(page.renderComeBackLater) {
             @govukButton(Button(content = Text(messages("linkText.saveAndComeBackLater")),
-                    inputType=Some("submit"),
-                    attributes = Map("formaction" -> uk.gov.hmrc.gform.gform.routes.FormController.updateFormData(formTemplate._id, maybeAccessCode, sectionNumber, fastForward, SaveAndExit).path),
-                    classes="govuk-button--secondary"
-                )
+                inputType = Some("submit"),
+                attributes = Map("form" -> "gf-form", "formaction" -> uk.gov.hmrc.gform.gform.routes.FormController.updateFormData(formTemplate._id, maybeAccessCode, sectionNumber, fastForward, SaveAndExit).path),
```
josef-vlach, line 147: Can we undo this change and move it back to gf-form form?

app/uk/gov/hmrc/gform/views/form/form.scala.html
```diff
@@ -124,25 +126,32 @@
...
     @* this is here for the sole purpose of making it the default submit button, when fields of group exist *@
-    @displaySaveContinueButton("defaultsink", Map("tabindex" -> "-1", "aria-hidden" -> "true"))
+    @displaySaveContinueButton("defaultsink", Map("tabindex" -> "-1", "aria-hidden" -> "true", "form" -> "gf-form"))
 
-      @{page.snippets}
+    @if(page.upscanData.isDefined){
+        @formWithoutCSRF(action = Call("POST", page.upscanData.get.url), args = 'class -> "js-form", 'autocomplete -> "on", 'novalidate -> "novalidate", 'id -> "gf-upscan"){
+            @{page.upscanData.get.snippets}
+        }
+    }
```
josef-vlach, line 137:
```suggestion
  @page.upscanData.map { upscanData =>
      @formWithoutCSRF(action = Call("POST", upscanData.url), args = 'class -> "js-form", 'autocomplete -> "on", 'novalidate -> "novalidate", 'id -> "gf-upscan"){
        @{upscanData.snippets}
      }
    }
```

As a rule of thumb, never use `.get` on Option type, there is always better way.

app/uk/gov/hmrc/gform/views/form/form.scala.html
```diff
@@ -124,25 +126,32 @@
       }
     }
 
-    @formWithoutCSRF(action = page.formAction, args = 'class -> "js-form", 'autocomplete -> "on", 'novalidate -> "novalidate", 'id -> "gf-form") {
 
     @* this is here for the sole purpose of making it the default submit button, when fields of group exist *@
-    @displaySaveContinueButton("defaultsink", Map("tabindex" -> "-1", "aria-hidden" -> "true"))
+    @displaySaveContinueButton("defaultsink", Map("tabindex" -> "-1", "aria-hidden" -> "true", "form" -> "gf-form"))
```
josef-vlach, line 131: Can we undo this change and move it back to gf-form form?

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -833,31 +871,31 @@ class SectionRenderingService(
...
-                  case (name, value) =>
-                    html.form.snippets.hidden(name, value)
-                }
                 val fileUploadName = "file"
-                val attributes = Map("upscan" -> "upscan")
+                val otherAttributes =
+                  if (upscanData.isDefined) Map("form" -> upscanData.get.formMetaData.htmlId)
+                  else Map.empty[String, String]
+
+                val formActionUrl = upscanData.map(_.url).getOrElse("")
```
josef-vlach, line 885: Calling `.get` on value of `Option` type is bad and please don't do it.

In this case when there is no `upscanData` available what this implementation would result in is a broken form. 

Right thing to do here is to enforce presence of `upscanData`, in case `upscanData` are not available there is bug in our code and there is nothing better we can do than to throw and exception.

```
case FileUploadProvider.Upscan =>
  upscanData match {
    case None => throw new IllegalArgumentException(TODO provide nice error message with as much info as possible)
    case Some(upscanData) =>
      val fileUploadName = "file"
      ...
  }

josef-vlach (changes requested):
Almost there 👍

app/uk/gov/hmrc/gform/views/form/form.scala.html
DursunGundogan: Done

app/uk/gov/hmrc/gform/views/form/form.scala.html
DursunGundogan: Done

app/uk/gov/hmrc/gform/views/form/form.scala.html
DursunGundogan: Done

josef-vlach (approved):
Good job!

PR #1362, opened 2021-11-18
GFORMS-1550 - Improve Upscan error messaging where file extension is incorrect

DursunGundogan:


public/javascripts/gformFileUpload.js
```diff
@@ -301,7 +297,7 @@
             return checkConfirmation(formTemplateId, key);
           }, onError)
           .then(function(response) {
-            if(response == "") {
+            if(response === undefined) {
```
josef-vlach, line 300: Under what condition is response `undefined`?

DursunGundogan: If the file is uploaded successfully, the `check` method returns `NoContent`. it means `response` is undefined.
https://github.com/hmrc/gform-frontend/blob/dg/GFORMS-1550/app/uk/gov/hmrc/gform/upscan/UpscanController.scala#L117

josef-vlach (approved):

PR #1372, opened 2021-11-28
testEmailForTestAutomation

ratheeshsm:


josef-vlach:
We can't have this go to the prod. Why don't you specify this param as java process parameter when running automation?

PR #1376, opened 2021-12-03
GFORMS-1596 - Make sort code one field in bank account details request

DursunGundogan:


josef-vlach (commented):
It looks okish.

I'm just little concerned that we accept `£#12%^34*&56><` as valid sort code (which is apparently ok) but I doubt that we should pass it to DMS in this weird shape.

Also when we are replaying sort code back to the user I think we should normalise it to `12-34-56` format?

Also we need to rebase this branch, sorry about that.

app/uk/gov/hmrc/gform/validation/ValidationResult.scala
```diff
@@ -63,6 +63,12 @@ case class ValidationResult(
       case (_, FieldOk(formComponent, cv)) if formComponent.isReferenceNumber =>
         val cvUpd: String = cv.replace(" ", "")
         FieldOk(formComponent, cvUpd)
+      case (_, FieldOk(formComponent, cv)) if formComponent.isSortCode =>
+        val cvUpd: String = cv
+          .replaceAll("[^0-9]", "")
+          .grouped(2)
+          .mkString("-")
```
josef-vlach, line 70: Saving sortcode with dashes would break our bank account check calls. We need to store sortcode as 6 digits and then format it when we want to replay it back to the user.

josef-vlach (changes requested):

josef-vlach (approved):

josef-vlach (approved):

PR #1378, opened 2021-12-08
GFORMS-1591 - Fix issue of resolving an email address for sending the confirmation email

DursunGundogan:


app/uk/gov/hmrc/gform/gformbackend/GformBackEndService.scala
```diff
@@ -189,6 +190,16 @@ class GformBackEndService(
...
+      maybeEmailAddress <- cache.formTemplate.emailExpr match {
+                             case Some(expr) =>
+                               formModelOptics.formModelVisibilityOptics
+                                 .evalAndApplyTypeInfoFirst(expr)
+                                 .expressionResult match {
+                                 case StringResult(str) => Future.successful(Some(str))
+                                 case _                 => Future.successful(None)
+                               }
+                             case None => Future.successful(None)
+                           }
```
josef-vlach, line 202: Can we simplify this to:

```suggestion
maybeEmailAddress <- cache.formTemplate.emailExpr.traverse { expr =>
  val email = formModelOptics.formModelVisibilityOptics
    .evalAndApplyTypeInfoFirst(expr)
    .stringRepresentation
  Future.successful(email)
}
```

(Using `import cats.syntax.all._`)

Rationale why to use `.stringRepresentation`: If email expression is for example address reference, it would be better to see exception (once we try to send the email) that address data are not valid email rather then whatever fallback error would be when we pretend that `emailExpr` was `None`

josef-vlach (changes requested):

josef-vlach (approved):

PR #1389, opened 2022-01-05
GFORMS-1607 - Introduce new expression size - return the number of ch…

DursunGundogan:
…oices in Add to List

### DIFF
--- a/app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
+++ b/app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
@@ -148,6 +148,21 @@ case class EvaluationResults(
     }
   }
 
+  private def evalSize(formComponentId: FormComponentId, recData: RecData[SourceOrigin.OutOfDate], index: Int) = {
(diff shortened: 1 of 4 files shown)


app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -337,6 +353,12 @@ case class EvaluationResults(
             } yield result).getOrElse("")
           )
         )
+      case Size(formComponentId, index) =>
+        nonEmpty(
+          StringResult(
+            evalSize(formComponentId, recData, index).stringRepresentation(typeInfo, evaluationContext.messages)
+          )
+        )
```
josef-vlach, line 361: This should be simply

```
evalSize(formComponentId, recData, index)
```

I don't see any reason for turning it into `StringResult`

josef-vlach (changes requested):
Looks good, only one small change needed.

josef-vlach (approved):

PR #1391, opened 2022-01-10
Revert: GFORMS-1583 - Fix styling issue on referrerConfig.exitMessage property

DursunGundogan:


josef-vlach:
This looks just like revert of [GFORMS-1583 - Fix styling issue on referrerConfig.exitMessage property](https://github.com/hmrc/gform-frontend/commit/e2a7594d97cab0b053432c0d755a08ba7bc69f79)
If I look correctly.

Does it mean GFORMS-1583 is being reintroduced?

Also could we just to revert said commit to make clear what is happening?

PR #1394, opened 2022-01-11
GFORMS-1640 - user is getting Bad request(400) after check summary pa…

DursunGundogan:
…ge when user try to use special character 'ü' in the ToR Form

### DIFF
--- a/app/uk/gov/hmrc/gform/gformbackend/GformConnector.scala
+++ b/app/uk/gov/hmrc/gform/gformbackend/GformConnector.scala
@@ -25,6 +25,7 @@ import cats.syntax.applicative._
 import cats.syntax.eq._
 import cats.syntax.functor._
 import cats.syntax.show._
+import org.apache.commons.codec.binary.Base64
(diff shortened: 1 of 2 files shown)


app/uk/gov/hmrc/gform/gformbackend/GformConnector.scala
```diff
@@ -250,7 +252,7 @@ class GformConnector(ws: WSHttp, baseUrl: String) {
     ws.POST[String, Submission](
       baseUrl + s"/forms/${formId.value}/${formTemplateId.value}/${envelopeId.value}/$noOfAttachments/createSubmission",
       "",
-      Seq(("customerId", customerId))
+      Seq(("customerId", Base64.encodeBase64String(customerId.getBytes(StandardCharsets.UTF_8))))
```
josef-vlach, line 255: Would viable option be to just html escape `customerId`? https://www.freeformatter.com/html-entities.html

josef-vlach (approved):

PR #1396, opened 2022-01-12
GFORMS-1614 - Accessibility Testing-Attribute “name” not allowed on e…

DursunGundogan:
…lement “meta” and Element “meta” is missing one or more of the following attributes: “itemprop”, “property”.


josef-vlach:
How is this different from what we had before https://github.com/hmrc/gform-frontend/pull/1395 was merged?
How is this fixing the problem?

DursunGundogan:
It should be outside of <script></script> tag. if we put in <script> tag, character counter is not working in text-area component

josef-vlach (approved):

PR #1399, opened 2022-01-14
GFORMS-1622 - Validators won't validate against validifs only

DursunGundogan:


app/uk/gov/hmrc/gform/graph/Recalculation.scala
```diff
@@ -331,6 +357,25 @@ class Recalculation[F[_]: Monad, E](
...
+  )(implicit formModel: FormModel[Interim]): StateT[F, RecalculationState, Boolean] = {
+    val validIfs: List[ValidIf] = formModel.allValidIfs.filter(_._2.id === fcId).flatMap(_._1)
+
+    validIfs
+      .traverse(validIf =>
+        evaluateValidIf(evaluationResults, recData, retrievals, booleanExprResolver, evaluationContext) {
+          Some(validIf)
+        }
+      )
+      .map(_.foldLeft(false)(_ && _))
```
josef-vlach, line 376: Name of this method `isHiddenByValidIf` doesn't sound right. `validIf` cannot hide anything.

Also this line (376) will always evaluate to `false`, unless I'm missing something.

I think (and I may be wrong) all we want is 
```
formModel.allValidIfs.traverse(validIf =>
  evaluateValidIf(evaluationResults, recData, retrievals, booleanExprResolver, evaluationContext) {
    Some(validIf)
  }
)
```

And combine it with result of 
```
graphLayer.foldMapM {
  ...
}
 ```
 
 
Also `evaluateValidIf` and `evaluateIncludeIf` are almost the same, we should refactor common code.

**Review by josef-vlach — CHANGES_REQUESTED** [review]

File: app/uk/gov/hmrc/gform/graph/Recalculation.scala [hunk 2]
```diff
@@ -166,9 +166,17 @@ class Recalculation[F[_]: Monad, E](
 
           noStateChange(evResult + (expr, exprResult))
       }
+
+      val validIfsResult: StateT[F, RecalculationState, Unit] =
+        evalValidIfs(evResult, recData, retrievals, booleanExprResolver, evaluationContext)
+
+      (validIfsResult, graphLayerResult).mapN { case (_, result) =>
+        result
+      }
```
josef-vlach, line 175: It is quite confusing what the purpose of this code is and why it works.

We should add comment to explain what is going on. ie. that we are only interested in `ValidIf` with `In` expression and that any other validIf is being ignored etc.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/graph/Recalculation.scala
josef-vlach: Also perhaps we can simplify the code a bit too:

```
evalValidIfs(evResult, recData, retrievals, booleanExprResolver, evaluationContext).flatMap { _ =>
   graphLayerResult
}
```

josef-vlach (approved):

PR #1408, opened 2022-01-24
GFORMS-1590 - Fix usability issue of new gForm being started when inf…

DursunGundogan:
…light form has not been submitted


app/uk/gov/hmrc/gform/gform/CryptoUtil.scala
```diff
@@ -0,0 +1,34 @@
...
+      new java.math.BigInteger(
+        1,
+        java.security.MessageDigest
+          .getInstance("SHA-256")
+          .digest(
+            text
+              .getBytes("UTF-8")
+          )
+      )
+    )
```
josef-vlach, line 33: I think we can simplify this:

```suggestion
import java.nio.charset.StandardCharsets

def sha256Hash(text: String): String =
    java.security.MessageDigest
      .getInstance("SHA-256")
      .digest(
        text
          .getBytes(StandardCharsets.UTF_8)
      )
      .map("%02x".format(_))
      .mkString
```

app/uk/gov/hmrc/gform/sharedmodel/form/FormData.scala
```diff
@@ -30,6 +31,8 @@ case class FormData(fields: List[FormField]) {
 
   def ++(other: FormData): FormData =
     FormData((toData ++ other.toData).map { case (k, v) => FormField(k, v) }.toList)
+
+  def toHash: Option[String] = Some(CryptoUtil.sha256Hash(Json.toJson(FormData(fields)).toString()))
```
josef-vlach, line 35: I rather not call `.toString`. There `Json.stringify` which seems like a better implementation.

josef-vlach: I don't see any reason for this to return an `Option`

conf/app.routes
```diff
@@ -37,7 +37,7 @@ POST          /form/addtolist/remove/:formTemplateId/:maybeAccessCode/:sectionNu
...
 GET           /summary/:formTemplateId/                                                                          uk.gov.hmrc.gform.gform.SummaryController.summaryById(formTemplateId: FormTemplateId, a: Option[AccessCode])
 GET           /summary/:formTemplateId/:maybeAccessCode                                                          uk.gov.hmrc.gform.gform.SummaryController.summaryById(formTemplateId: FormTemplateId, maybeAccessCode: Option[AccessCode])
-POST          /summary/:formTemplateId/:maybeAccessCode                                                          uk.gov.hmrc.gform.gform.SummaryController.submit(formTemplateId: FormTemplateId, maybeAccessCode: Option[AccessCode], action: Direction)
+POST          /summary/:formTemplateId/:maybeAccessCode                                                          uk.gov.hmrc.gform.gform.SummaryController.submit(formTemplateId: FormTemplateId, maybeAccessCode: Option[AccessCode], action: Direction, refreshToken: Option[String])
```
josef-vlach, line 40: I don't see any reason why `refreshToken` should be an `Option`. Maybe I'm missing something?

app/uk/gov/hmrc/gform/sharedmodel/form/FormData.scala
josef-vlach: Also what about renaming `toHash` to `fingerprint` ?

josef-vlach (changes requested):
Couple of minor points. 

Also I'm not sure about calling this token a `refreshToken`. It is more like `formDataFingeprint`.

PR #1417, opened 2022-02-03
GFORMS-1670 - The emailConfirmation property is being ignored in comp…

DursunGundogan:
…osite auth

### DIFF
--- a/app/uk/gov/hmrc/gform/gform/EmailAuthController.scala
+++ b/app/uk/gov/hmrc/gform/gform/EmailAuthController.scala
@@ -34,12 +34,11 @@ import uk.gov.hmrc.gform.gform.SessionUtil.jsonFromSession
 import uk.gov.hmrc.gform.gformbackend.GformConnector
 import uk.gov.hmrc.gform.models.EmailId
 import uk.gov.hmrc.gform.models.optics.DataOrigin
-import uk.gov.hmrc.gform.sharedmodel.formtemplate.FormTemplateWithRedirects
+import uk.gov.hmrc.gform.sharedmodel.formtemplate.{ AuthConfig, Composite, EmailAuthConfig, FormTemplateId, FormTemplateWithRedirects, HasEmailConfirmation }
 import uk.gov.hmrc.gform.sharedmodel.{ LangADT, LocalisedString }
 import uk.gov.hmrc.gform.sharedmodel.email.{ ConfirmationCodeWithEmailService, EmailConfirmationCode }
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/gform/gform/EmailAuthController.scala
```diff
@@ -344,6 +350,21 @@ class EmailAuthController(
               MarkDownUtil.markDownParser(emailConfirmation)
             )
           ).pure[Future]
+        case Composite(configs) if configs.exists(_.isEmailConfirmation) =>
```
josef-vlach, line 353: There is a bit of duplication here. What about introducing `HasEmailConfirmation` extractor so instead of these two cases

```
case EmailAuthConfig(_, _, _, Some(emailConfirmation))
case Composite(configs) if configs.exists(_.isEmailConfirmation)
```

We would need to do only:

```
case HasEmailConfirmation(emailConfirmation)
```

?

app/uk/gov/hmrc/gform/gform/EmailAuthController.scala
```diff
@@ -315,6 +315,12 @@ class EmailAuthController(
                     ).addingToSession(
                       EMAIL_AUTH_DETAILS_SESSION_KEY -> confirmedEmailAuthDetailsStr
                     )
+                  case Composite(configs) if configs.exists(_.isEmailConfirmation) =>
```
josef-vlach, line 318: `HasEmailConfirmation` would find a use also here

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/AuthConfig.scala
```diff
@@ -135,6 +141,16 @@ object HasEnrolmentSection {
     }
 }
 
+object HasEmailConfirmation {
+  def unapply(ac: AuthConfig): Option[LocalisedString] =
+    ac match {
+      case EmailAuthConfig(_, _, _, Some(emailConfirmation)) => Some(emailConfirmation)
+      case Composite(configs) if configs.exists(_.isEmailConfirmation) =>
```
josef-vlach, line 148: I think this `if configs.exists(_.isEmailConfirmation)` is not needed now and so `def isEmailConfirmation` can also be removed?

josef-vlach (changes requested):

josef-vlach (approved):

PR #1418, opened 2022-02-09
GFORMS-1669 - Ensure all new users trying to access v1 url are redire…

DursunGundogan:
…cted to latest url

### DIFF
--- a/app/uk/gov/hmrc/gform/gform/NewFormController.scala
+++ b/app/uk/gov/hmrc/gform/gform/NewFormController.scala
@@ -249,7 +249,7 @@ class NewFormController(
     for {
       formTemplate <- gformConnector.getFormTemplate(formTemplateId)
       formIdData   <- Future.successful(FormIdData.Plain(UserId(retrievals), formTemplate.formTemplate._id))
-      maybeForm    <- gformConnector.maybeForm(formIdData, formTemplate.formTemplate)
+      maybeForm    <- gformConnector.maybeForm(formIdData)
       res <- maybeForm match {
                case Some(_) => Future.successful(formTemplate.formTemplate)
                case None =>
--- a/app/uk/gov/hmrc/gform/gform/NewFormController.scala
+++ b/app/uk/gov/hmrc/gform/gform/NewFormController.scala
@@ -213,7 +213,7 @@ class NewFormController(
         val queryParams: QueryParams = QueryParams.fromRequest(request)
 
         for {
-          formTemplate <- getLatestFormTemplate(formTemplateId, cache.retrievals)
+          formTemplate <- getLatestFormTemplate(formTemplateId)
           formIdData   <- Future.successful(FormIdData.Plain(UserId(cache.retrievals), formTemplate._id))
           res <-
             handleForm(formIdData, formTemplate)(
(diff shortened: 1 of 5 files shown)


josef-vlach (changes requested):
This is undoubtedly working solution, but it is not the way how we should solve this problem.

We should use data in `formRedirect` collection to determine latest `FormTemplateId`.

app/uk/gov/hmrc/gform/gform/NewFormController.scala
```diff
@@ -212,22 +212,54 @@ class NewFormController(
...
+                case _ =>
+                  auditService.sendFormResumeEvent(form, cache.retrievals)
+                  val continueFormPage = new ContinueFormPage(formTemplate, choice)
+                  Ok(continue_form_page(frontendAppConfig, continueFormPage)).pure[Future]
+              }
+            }
+        } yield res
     }
 
+  private def getLatestFormTemplate(
```
josef-vlach, line 245: This method doesn't do what name suggest. I'm not sure the logic of this method is correct.
Why we simply don't get latest `FormTemplateId`  here? Why all this form existence checking?

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/gform/NewFormController.scala
DursunGundogan: According to AC4 [GFORMS-1669](https://jira.tools.tax.service.gov.uk/browse/GFORMS-1669) if there is a saved form for any previous version, the form will be opened. I think we need to check whether a form exists or not for all previous versions.

josef-vlach: I agree with AC4. But for simplicity sake we should get latest formTemplate and rely on `GformConnector.maybeForm` to find any potential old form.

In other words we should go from latest (youngest) to the oldest formTemplate, rather than from oldest to youngest, as this solution do.

app/uk/gov/hmrc/gform/gform/NewFormController.scala
```diff
@@ -212,22 +212,48 @@ class NewFormController(
...
+    formTemplateId: FormTemplateId
+  )(implicit hc: HeaderCarrier): Future[FormTemplate] =
+    for {
+      formTemplate <- gformConnector.getFormTemplate(formTemplateId)
+      res <-
+        formTemplate.redirect match {
+          case Some(ft) => getLatestFormTemplate(ft)
+          case None     => Future.successful(formTemplate.formTemplate)
```
josef-vlach, line 253: Can we do this with just one call to the backend?

We should add new method to the `gformConnector`: 
```
def getLatestFormTemplate(
  formTemplateId: FormTemplateId
)
```

And we should implement it on the backend in terms of `formRedirect` collection.

We have all the data ready for us to be processed.
```
> db.formRedirect.find({}).pretty();
{
	"_id" : "version-test-1",
	"redirect" : "version-test",
	"createdAt" : "2022-02-14T13:59:00.232754Z",
	"updatedAt" : "2022-02-14T13:59:00.232756Z"
}
{
	"_id" : "version-test-0",
	"redirect" : "version-test-1",
	"createdAt" : "2022-02-14T13:59:39.306877Z",
	"updatedAt" : "2022-02-14T13:59:39.306879Z"
}
```

josef-vlach (changes requested):

josef-vlach (approved):

PR #1420, opened 2022-02-10
GFORMS-1657 - Add new component 'taxPeriodDate' to allow a date without a day

rhubner:


app/uk/gov/hmrc/gform/validation/TaxPeriodDateValidator.scala
```diff
@@ -0,0 +1,115 @@
...
+  ): ValidatedType[Unit] = {
+
+    case class ModelComponentIdValue(modelComponentId: ModelComponentId, value: Option[String])
+
+    val atomsWithValues: List[ModelComponentIdValue] = formComponent.multiValueId.atomsModelComponentIds.map(m =>
+      ModelComponentIdValue(m, formModelVisibilityOptics.data.one(m))
+    )
+
+    val validatedResult = if (formComponent.mandatory) {
+      atomsWithValues.map { mcv =>
```
josef-vlach, line 95: This `map` can be changed to `foldMap` so the results of the lambda are immediately combined with theirs `Monoid` instance.

josef-vlach (changes requested):
Couple of small changes

josef-vlach (approved):

PR #1430, opened 2022-02-25
GFORMS-1677 - DAC - Unclear relations - The description property in A…

DursunGundogan:
…TL should be marked as ordered list on summary page


app/uk/gov/hmrc/gform/models/ExpandUtils.scala
```diff
@@ -50,7 +50,7 @@ object ExpandUtils {
 
   def expandSmartString(smartString: SmartString, index: Int, ids: List[FormComponentId]): SmartString = {
     val interpolations: List[Expr] = smartString.interpolations.map(expr => ExprUpdater(expr, index, ids))
-    smartString.copy(interpolations = interpolations).replace("$n", index.toString)
+    smartString.copy(interpolations = interpolations).replace("$n", "")
```
josef-vlach, line 53: This would break `$n` functionality, we cannot do that.

Unfortunately I haven't any suggestion how to fix this issue.

josef-vlach (changes requested):

josef-vlach (approved):

PR #1444, opened 2022-03-04
GFORMS-1691 - Add postcode validation to address component

lxol:


app/uk/gov/hmrc/gform/validation/AddressValidation.scala
```diff
@@ -186,4 +187,18 @@ class AddressValidation[D <: DataOrigin](implicit messages: Messages, sse: Smart
...
+
+  private def checkPostcode(
+    fieldValue: FormComponent,
+    atomicFcId: ModelComponentId.Atomic,
+    messageKey: String
+  )(
+    xs: Seq[String]
+  ): ValidatedType[Unit] =
+    xs.filterNot(_.isEmpty) match {
+      case value :: Nil if !PostcodeLookupValidation.checkPostcode(value) =>
```
lxol, line 199: Should we extract validation logic to some utility class?

josef-vlach: `lengthLimitValidation ` and `checkPostcode` differs only on predicate we should refactor that to have single implementation, but private method sounds good enough. But if you can see that this pattern is repeated elsewhere than fell free to do broader refactoring using utility class

josef-vlach (changes requested):

PR #1447, opened 2022-03-07
GFORMS-1668 - Fix issue with how captions are implemented in gForms u…

DursunGundogan:
…sing progressIndicator


app/uk/gov/hmrc/gform/views/form/form.scala.html
```diff
@@ -114,7 +114,7 @@
     } else {
       @* This is here to support caption in case of isPageHading case *@
       @page.progressIndicator.fold(HtmlFormat.empty){ caption =>
-        <p class="govuk-!-margin-bottom-0 govuk-caption-l hmrc-caption-xl"><span class="govuk-visually-hidden">@messages("this.section.is")</span>@caption</p>
+          <span class="govuk-caption-l">@caption</span>
```
josef-vlach, line 117: I think this is breaking accessibility. Can you point me to the source where this change was taken from?

josef-vlach: This feels like going in the circle which is wrong. 
GOV UK Design system is a standard not an implementation. Implementation we are using is this: https://github.com/hmrc/play-frontend-hmrc
This library evolved a lot since the last time we tried to fully utilise it and we need to catch up with it now. End goal is not to write html tags ourselves, but use this library to do it for us.

lxol (approved):

lxol (approved):

PR #1452, opened 2022-03-11
GFORMS-1644: fix - Direction Back is not supported here

lxol:


josef-vlach (commented):
This looks like it is on right path. but I wonder why simple don't change only `request.uri` for `.../submissions/new-form/...` on this line: https://github.com/hmrc/gform-frontend/blob/main/app/uk/gov/hmrc/gform/auth/models/RecoverAuthResult.scala#L53
?

lxol:
the problem there is that templateId is not available there. I had to wire it to the recovery method

josef-vlach:
Can you summon it like this:
```
val formTemplateWithRedirect: FormTemplateWithRedirects = request.attrs(FormTemplateKey)
```
?

lxol:
Do I need to handle `NoSuchElementException` if attribute is not in the request's header?
I can't see that it is handled anywhere in the code base.

app/uk/gov/hmrc/gform/auth/models/RecoverAuthResult.scala
```diff
@@ -48,12 +50,24 @@ object RecoverAuthResult {
...
     request: Request[AnyContent],
     appConfig: AppConfig
-  ): PartialFunction[Throwable, AuthResult] = { case _: NoActiveSession =>
-    logger.debug("No Active Session")
-    val continueUrl = java.net.URLEncoder.encode(appConfig.`gform-frontend-base-url` + request.uri, "UTF-8")
-    val ggLoginUrl = appConfig.`government-gateway-sign-in-url`
-    val url = s"$ggLoginUrl?continue=$continueUrl"
-    AuthRedirectFlashingFormName(url)
+  ): PartialFunction[Throwable, AuthResult] = {
+    case _: NoActiveSession if request.method == "POST" =>
```
josef-vlach, line 54: Let's make this the same for all requests, no need to handle only POST differently.

app/uk/gov/hmrc/gform/auth/models/RecoverAuthResult.scala
```diff
@@ -48,12 +50,24 @@ object RecoverAuthResult {
...
+      val formTemplateWithRedirect = request.attrs(FormTemplateKey)
+      val formTemplate = formTemplateWithRedirect.formTemplate
+      logger.debug("No Active Session")
+      val continueUrl = URLEncoder.encode(
+        s"${appConfig.`gform-frontend-base-url`}/submissions/new-form/${formTemplate._id.value}",
```
josef-vlach, line 59: Let's not hardcode `/submissions/new-form/` here. 
We can use reverse route like `routes.NewFormController.dashboard(...)` instead.

josef-vlach (changes requested):

josef-vlach (approved):

PR #1454, opened 2022-03-16
GFORMS-1695 - Fix issue with implementing composite auth - users with…

DursunGundogan:
… saved data in old form must verify email twice


josef-vlach:
I've been testing this change but it seems to be more broken than it was before.

PR #1458, opened 2022-03-21
GFORMS-791: Enhance choice component to allow for a text divider

lxol:


app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -1218,25 +1233,43 @@ class SectionRenderingService(
...
+              CheckboxItem(
+                id = Some(formComponent.id.value + index),
+                value = index.toString,
+                content = content.Text(option.value),
+                checked = isChecked(index),
+                conditionalHtml = helpTextHtml(maybeHelpText),
+                attributes = dataLabelAttribute(option),
+                hint = maybeHint
+              )
+            }
```
josef-vlach, line 1259: There is quite a duplication here. What about to use 
```suggestion
          val item = CheckboxItem(
            id = Some(formComponent.id.value + index),
            value = index.toString,
            content = content.Text(option.value),
            checked = isChecked(index),
            conditionalHtml = helpTextHtml(maybeHelpText),
            attributes = dataLabelAttribute(option),
            hint = maybeHint
          )
          if (noneChoice.contains(index + 1)) {
            item.copy(behaviour = Some(ExclusiveCheckbox))
          } else {
            item
          }
```

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -1218,25 +1233,43 @@ class SectionRenderingService(
...
-            content = content.Text(option.value),
-            checked = isChecked(index),
-            conditionalHtml = helpTextHtml(maybeHelpText),
-            attributes = dataLabelAttribute(option),
-            hint = maybeHint
-          )
-        }
+        val items = dividerPosition.foldLeft(optionsWithHintAndHelpText.zipWithIndex.map {
+          case ((option, maybeHint, maybeHelpText), index) =>
+            if (noneChoice == Some(index + 1)) {
```
josef-vlach, line 1238: Use more idiomatic `contains`  rather than `... == Some(...)`

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -1218,25 +1233,43 @@ class SectionRenderingService(
...
-            id = Some(formComponent.id.value + index),
-            value = index.toString,
-            content = content.Text(option.value),
-            checked = isChecked(index),
-            conditionalHtml = helpTextHtml(maybeHelpText),
-            attributes = dataLabelAttribute(option),
-            hint = maybeHint
-          )
-        }
+        val items = dividerPosition.foldLeft(optionsWithHintAndHelpText.zipWithIndex.map {
```
josef-vlach, line 1236: Can we un-inline that value passed to `foldLeft` for better readability?

josef-vlach (changes requested):

PR #1459, opened 2022-03-23
GFORMS-1712 - An invalid XML character (Unicode: 0x2) was found in th…

DursunGundogan:
…e element content of the document.Can't load the XML resource (using TRaX transformer). org.xml.sax.SAXParseException


app/uk/gov/hmrc/gform/controllers/helpers/FormDataHelpers.scala
```diff
@@ -52,22 +53,24 @@ object FormDataHelpers {
...
-                s"Found invisible characters in field $field. " +
-                  s"Matches are [${matches
-                    .map { case (m, count) =>
-                      s"${getUnicode(m)}:${getDesc(m)}($count)"
-                    }
-                    .mkString(", ")}]"
-              )
-              replaceInvisibleChars(value).trim
+          values
+            .map(StringEscapeUtils.unescapeHtml4)
```
josef-vlach, line 57: Why is this needed?

app/uk/gov/hmrc/gform/controllers/helpers/InvisibleCharsHelper.scala
```diff
@@ -59,7 +59,38 @@ object InvisibleCharsHelper {
...
+    UnicodeDescMapping('\u0003', "End Of Text", ""),
+    UnicodeDescMapping('\u0004', "End Of Transmission", ""),
+    UnicodeDescMapping('\u0005', "Enquiry", ""),
+    UnicodeDescMapping('\u0006', "Acknowledge", ""),
+    UnicodeDescMapping('\u0007', "Bell", ""),
+    UnicodeDescMapping('\u0008', "Backspace", ""),
+    UnicodeDescMapping('\u0009', "Character Tabulation", ""),
+    UnicodeDescMapping('\u000B', "Line Tabulation", ""),
+    UnicodeDescMapping('\u000C', "Form Feed (FF)", ""),
+    UnicodeDescMapping('\u000D', "Carriage Return (CR)", ""),
```
josef-vlach, line 75: I think we want to accept Carriage Return

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/controllers/helpers/FormDataHelpers.scala
DursunGundogan: We need to convert it from the `HTML Entity` to Unicode. For example; when a user enters  `&#2;`, it needs to be converted to Unicode to find in the INVISIBLE_UNICODE_DESC_MAPPING.

DursunGundogan: It needs to be converted from HTML Entity(like `&#2;`) to Unicode(\u0002) to find it in the INVISIBLE_UNICODE_DESC_MAPPING

josef-vlach: What if I enter it as unicode char?

DursunGundogan: It replaces both of them. unit test is [here](https://github.com/hmrc/gform-frontend/pull/1459/files#diff-9266775e40fad540ff83cfa8089a9689143a4302a83129759182c06bd1ad8d8eR251)

josef-vlach: Oh, now I see it. It's clever.

josef-vlach (approved):

josef-vlach:
If I enter `&#2;` into a text field it doesn't seem to matter to pdf generator.

app/uk/gov/hmrc/gform/controllers/helpers/FormDataHelpers.scala
```diff
@@ -39,6 +40,9 @@ object FormDataHelpers {
 
   private val logger = LoggerFactory.getLogger(getClass)
 
+  /** This filters HTML Entities (decimal/hex) or unicode from request body when entered.
+    * The pdf generator does not accept these characters.
+    */
```
josef-vlach, line 45: This comment is not correct. Problem is when we try to render (as markdown) SmartString which contains expression which refers to user input. Example is `summaryDescription` json field.

josef-vlach (approved):

PR #1461, opened 2022-03-24
GFORMS-791: add noneChoiceError

lxol:


app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
```diff
@@ -554,6 +554,23 @@ object ComponentValidator {
...
 
+  def validateChoiceNoneError[D <: DataOrigin](
+    fieldValue: FormComponent,
+    noneChoice: Int,
+    error: LocalisedString
+  )(
+    formModelVisibilityOptics: FormModelVisibilityOptics[D]
+  )(implicit l: LangADT): ValidatedType[Unit] = {
+    val choiceValue = formModelVisibilityOptics.data
+      .get(fieldValue.modelComponentId)
```
josef-vlach, line 565: Please call `many` not `get`. 

`many` is for components which can produce multiple answers, which in our case is only choice

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
lxol: ok, I got inspiration from 
https://github.com/hmrc/gform-frontend/blob/e7d469c09104c79f3039b75270ff3affe5fe6809/app/uk/gov/hmrc/gform/validation/ComponentValidator.scala#L548
should we update it here as well?

app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
```diff
@@ -554,6 +554,23 @@ object ComponentValidator {
...
+    fieldValue: FormComponent,
+    noneChoice: Int,
+    error: LocalisedString
+  )(
+    formModelVisibilityOptics: FormModelVisibilityOptics[D]
+  )(implicit l: LangADT): ValidatedType[Unit] = {
+    val choiceValue = formModelVisibilityOptics.data
+      .many(fieldValue.modelComponentId)
+      .toSeq
+      .flatMap(_.toSeq)
```
josef-vlach, line 567:
```suggestion
      .flatten
```
One suggestion

app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
josef-vlach: Yes, we should change it there as well.

app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
lxol: done

josef-vlach (approved):

PR #1476, opened 2022-04-11
GFORMS-1721 - Add message onto summary pdf to indicate that it is draft

DursunGundogan:


app/uk/gov/hmrc/gform/gform/SummaryController.scala
```diff
@@ -256,6 +256,16 @@ class SummaryController(
       maybeAccessCode,
       OperationWithForm.DownloadSummaryPdf
     ) { implicit request => implicit l => cache => implicit sse => formModelOptics =>
+      val maybeWatermark = cache.formTemplate.summarySection.default.flatMap { isDefault =>
```
lxol, line 259:
```suggestion
          val maybeWatermark = cache.formTemplate.summarySection.default.collect { case true =>
        cache.formTemplate.formCategory match {
          case HMRCReturnForm => Messages("summary.pdf.formCategory.return")
          case HMRCClaimForm  => Messages("summary.pdf.formCategory.claim")
          case _              => Messages("summary.pdf.formCategory.form")
        }
      }

```

DursunGundogan: Done

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/SummarySection.scala
```diff
@@ -26,7 +26,8 @@ case class SummarySection(
   header: SmartString,
   footer: SmartString,
   continueLabel: Option[SmartString],
-  fields: Option[NonEmptyList[FormComponent]]
+  fields: Option[NonEmptyList[FormComponent]],
+  default: Option[Boolean] = None
```
josef-vlach, line 30: Why is this field called `default` and why it is optional `Boolean`?

DursunGundogan: If the summary section is not present,  the backend generates a default summary section. It needs to know which is default or not. Right, it can be represented by only Boolean without Option.

DursunGundogan: but if we do it as Boolean, we need to put it in JSON.

DursunGundogan: but if we do it as a boolean, the `default` attribute will be mandatory in JSON. I think it should be optional to ignore in Json

josef-vlach (approved):

PR #1483, opened 2022-04-21
GFORMS-1724 - Update format 'ukVrn' to allow 'XI' prefix or no prefix…

josef-vlach:
… to VAT number


app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
```diff
@@ -388,7 +388,12 @@ object ComponentValidator {
       case tooShort if tooShort.length < 7 =>
         val vars: List[String] = 7.toString :: Nil
         validationFailure(fieldValue, genericErrorMinLength, Some(vars))
-      case Standard()   => validationSuccess
+      case Standard(_, s) =>
```
lxol, line 391: should it be 
```
case Standard(_, s) if VatReferenceChecker.isValid(s) => validationSuccess
```
and the failure case will be handled by `case _ => ...` line 400

josef-vlach: Good catch 👍

lxol (approved):

PR #1485, opened 2022-04-22
GFORMS-1722: Allow conf of width of summary list section on CYA

lxol:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/SummarySection.scala
```diff
@@ -49,5 +50,6 @@ case class SummarySection(
 }
 
 object SummarySection extends JsonUtils {
-  implicit val format: OFormat[SummarySection] = Json.format[SummarySection]
+  val jsWithDefaults = Json.using[Json.WithDefaultValues]
+  implicit val format: OFormat[SummarySection] = jsWithDefaults.format[SummarySection]
```
josef-vlach, line 54: We should be using derived format on frontend.  Why is `Json.format[SummarySection]` not good enough?

app/uk/gov/hmrc/gform/views/summary/summary.scala.html
```diff
@@ -55,21 +57,37 @@
...
 
 
+    @twoThirdsMainContent {
     <h1 class="govuk-heading-l">
         @heading
     </h1>
     <div class="govuk-body govuk-!-margin-bottom-9">
 	  @header
     </div>
+    }
```
josef-vlach, line 88: Can we indent content of this `twoThirdsMainContent`?

app/uk/gov/hmrc/gform/views/summary/summary.scala.html
```diff
@@ -79,7 +97,10 @@ <h1 class="govuk-heading-l">
         }
 
       	@snippets
+    }
```
josef-vlach, line 100: Can we indent content of this `mainContent`?

app/uk/gov/hmrc/gform/views/summary/summary.scala.html
```diff
@@ -103,6 +124,7 @@ <h1 class="govuk-heading-l">
                 )
             }
         </div>
+        }
```
josef-vlach, line 127: Can we indent content of this `twoThirdsMainContent` properly?

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/SummarySection.scala
lxol: it is derived, but with support of default value of `displayWidth`

josef-vlach: But we don't need default value support, since this value will always be populated to us from the backend.

lxol: good point, I reverted to the original format

josef-vlach (approved):

PR #1490, opened 2022-04-27
GFORMS-1727 - Add a new expression that checks whether item is first …

DursunGundogan:
…item in Add To List

### DIFF
--- a/app/uk/gov/hmrc/gform/gform/FormController.scala
+++ b/app/uk/gov/hmrc/gform/gform/FormController.scala
@@ -358,9 +358,7 @@ class FormController(
       implicit request => implicit l => cache => _ => formModelOptics =>
         def processEditAddToList(processData: ProcessData, idx: Int, addToListId: AddToListId): Future[Result] = {
 
-          val brackets: BracketsWithSectionNumber[Visibility] =
-            processData.formModelOptics.formModelVisibilityOptics.formModel.brackets
-          val addToListIteration = brackets.addToListById(addToListId, idx)
+          val addToListIteration = processData.formModel.brackets.addToListById(addToListId, idx)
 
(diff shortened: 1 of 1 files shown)


app/uk/gov/hmrc/gform/gform/FormController.scala
```diff
@@ -358,11 +358,23 @@ class FormController(
       implicit request => implicit l => cache => _ => formModelOptics =>
         def processEditAddToList(processData: ProcessData, idx: Int, addToListId: AddToListId): Future[Result] = {
 
-          val addToListIteration =
-            processData.formModelOptics.formModelVisibilityOptics.formModel.brackets.addToListById(addToListId, idx)
+          val brackets: BracketsWithSectionNumber[Visibility] =
```
josef-vlach, line 361: I don't think switching to visibility model was a good decision.

Note that `"defaultPage"` can have an `"includeIf"` so in Visibility model `addToListIteration.secondSectionNumber` may not succeed.

app/uk/gov/hmrc/gform/gform/FormController.scala
```diff
@@ -358,11 +358,23 @@ class FormController(
...
+            processData.formModelOptics.formModelVisibilityOptics.formModel.brackets
+          val addToListIteration = brackets.addToListById(addToListId, idx)
 
           def defaultNavigation(): (SectionNumber, FastForward) = {
-            val firstAddToListPageSN: SectionNumber = addToListIteration.firstSectionNumber
+            val firstAddToListPageSN: SectionNumber = processData.formModel.brackets
```
josef-vlach, line 366: We should find `firstAddToListPageSN` in view model and then look to visibility model for same or bigger section number.

josef-vlach (approved):

PR #1494, opened 2022-05-03
GFORMS-1615 - Country type ahead / look up error state incorrectly su…

josef-vlach:
…ggests user entry


app/uk/gov/hmrc/gform/validation/FormFieldValidationResult.scala
```diff
@@ -76,6 +76,12 @@ trait FormFieldValidationResult {
 
   def formComponent: FormComponent
 
+  def forHtmlFieldId(htmlFieldId: HtmlFieldId): FormFieldValidationResult = this match {
```
lxol, line 79: should it be for**m**HtmlFieldId ?

lxol (approved):

PR #1504, opened 2022-05-12
GFORMS-1134 - Upscan: Introduce a property to define what file types …

josef-vlach:
…are allowed in a form.


josef-vlach:
We have been mixing upscan errors with our validation errors leading to some mess in the code. This is attempt to keep track of where the upscan error originates from.

lxol (approved):

PR #1508, opened 2022-05-18
GFORMS-1499: Lookup component: Error message update

lxol:


app/uk/gov/hmrc/gform/sharedmodel/LocalisedString.scala
```diff
@@ -23,6 +23,8 @@ case class LocalisedString(m: Map[LangADT, String]) {
   def value(implicit l: LangADT): String = m.getOrElse(l, m.getOrElse(LangADT.En, ""))
   def replace(toReplace: String, replaceWith: String): LocalisedString =
     copy(m = (m.map { case (lang, message) => (lang, message.replace(toReplace, replaceWith)) }))
+  def uncapitalize(): LocalisedString =
+    copy(m = m.map { case (lang, message) => (lang, message.splitAt(1) match { case (c, cs) => c.toLowerCase + cs }) })
```
josef-vlach, line 27: Why is it lowercasing only first char? Should it lowercase all text?

josef-vlach (changes requested):
This change breaks `overseasAddress` component. This component uses `lookup(country)` under the hood.

app/uk/gov/hmrc/gform/sharedmodel/LocalisedString.scala
lxol: I might over-complicated things but here is my thoughts
`shortName` of `label` will be used in the middle of the sentence "Select <shortName> from the list."
The form designer might have <short name like "Foo Bar Baz". So I decided to lower case only first letter to make it fit in the middle of the sentence. 
But yes, if we make the whole <shortName> lowercased it would make things easier.
What do you think?

josef-vlach: This doesn't look right:
```
No match for "North Korie". Select eU country of IOSS registration from the list.
```

lxol: yes, not sure what to do that. This is not a technical decision here I think.

app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
```diff
@@ -123,7 +123,11 @@ object ComponentValidator {
 
     def lookupError: ValidatedType[Unit] = {
       val vars: List[String] = lookupLabel.label :: Nil
-      validationFailure(fieldValue, genericErrorLookup, Some(vars))
+
+      val shortNameForErrorMessage =
+        SmartString(LocalisedString(Map(LangADT.En -> "a value", LangADT.Cy -> " werth")), Nil)
```
josef-vlach, line 128: Why don't put `"a value"` directly to the message?

conf/messages.cy
```diff
@@ -125,7 +125,7 @@ generic.governmentId.error.pattern=Rhaid nodi {0} yn y fformat cywir
 generic.governmentId.not.exist=Nid yw’r {0} a nodoch yn bodoli
 generic.error.sortCode=Nodwch god didoli dilys fel 123456
 generic.error.exactNumbers=Rhaid i {0} fod yn rhif {1} digid
-generic.error.lookup=Rhaid i {0} gyd-fynd ag un o’r awgrymiadau o’r rhestr
+generic.error.lookup=Dim cyfateb i {1}. Dewiswch {0} o'r rhestr.
```
josef-vlach, line 128: This `{1}` is not populated on `overseasAddress` component. This need fixing

app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
lxol: good point!

conf/messages
```diff
@@ -126,7 +126,7 @@ generic.governmentId.error.pattern={0} must be entered in the correct format
 generic.governmentId.not.exist=The {0} you entered does not exist
 generic.error.sortCode=Enter a valid sort code like 123456
 generic.error.exactNumbers={0} must be {1} numbers
-generic.error.lookup={0} must match one of the suggestions from the list
+generic.error.lookup=No match for "{1}". Select a value from the list.
```
josef-vlach, line 129: I've said it twice, I'll say it for third time:

This `{1}` is not populated on overseasAddress component. This need fixing

josef-vlach (changes requested):

conf/messages
lxol: done

josef-vlach (approved):

PR #1514, opened 2022-05-30
GFORMS-1744 - Allow choices to be displayed based on an includeIf expression

DursunGundogan:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/ComponentType.scala
```diff
@@ -331,7 +333,7 @@ object NoneChoice {
 
 case class Choice(
   `type`: ChoiceType,
-  options: NonEmptyList[OptionData],
+  options: List[OptionData],
```
josef-vlach, line 336: What will happen from user point of view if all choice are hidden?

DursunGundogan: It will display only the title of the component so it's a broken use case. Do you have any suggestions about it?

josef-vlach: That's what I'd expect. We need at least on option for the choice so we should put `NonEmptyList` back. And in case we can't get a `NonEmptyList` we will let it crash stating exactly the reason what happened in the exception.

josef-vlach (approved):

PR #1517, opened 2022-06-01
GFORMS-1714: fix contains

lxol:


app/uk/gov/hmrc/gform/eval/ExpressionResult.scala
```diff
@@ -126,8 +126,8 @@ sealed trait ExpressionResult extends Product with Serializable {
     case t: Invalid      => false
     case t: Hidden.type  => false
     case t: Empty.type   => false
-    case t: NumberResult => false
-    case t: StringResult => false
+    case t: NumberResult => identical(er)
+    case t: StringResult => identical(er)
```
josef-vlach, line 130: This is incorrect, why is this change being done? Example template?

lxol: This is to fix a failed template in the gform-acceptance
https://github.com/hmrc/gform-acceptance-tests/blob/f14a9d0175e2473cb2a5d767564403c2533959df/src/test/resources/jsonComponent/addtolist/addtolist-MiniCYAPage.json#L312

after GFORM-1714 was implented `Telephone2`, `country1` in the `includeIf` are Lists now. So we decided to use `contains` instead of `=`
but original implementation of `contains` for list 
https://github.com/hmrc/gform-frontend/blob/4c9cfbc7a655aeecbad855876787b47f3cf7e8ac/app/uk/gov/hmrc/gform/eval/ExpressionResult.scala#L139

It runs contains on each element of the list which in our case if StringResult or NumberResult. 
So instead of reimplenting `contains` for list I decided to alter implementation of `contains` for NumberResult and SringResult
Alternatively I think we can implement it like

```
case t: ListResult      => t.list.exists(_.identical(er))
```

josef-vlach: Isn't this fixed by #1516?

lxol: no, it is different

lxol: well not completely different. That change fixes the top level references to the ATL identifiers. They became Lists after the change. But `contains` on that list wont't work because `t.list.exists(_.contains(er))` and the function in `exists` always returns `false`

josef-vlach: I can't comprehend what you are saying. 

This looks correct to me:
```
t.list.exists(_.contains(er))
```

What you mean by
> the function in `exists` always returns `false`

BTW. I'm trying to test this and I'm failing to reproduce any case when it would not work as expected

lxol:
> What you mean by
>the function in exists always returns false

sorry I meant `contains` not `exists`

the behavior that I wanted to fix is illustrated in this snippet
```
import uk.gov.hmrc.gform.eval.ExpressionResult._
import uk.gov.hmrc.gform.eval._

val lr =ListResult(List(NumberResult(1), NumberResult(2)))
// lr: ListResult = ListResult(List(NumberResult(1), NumberResult(2)))
lr.contains(NumberResult(2))
// res0: Boolean = true
lr.contains(NumberResult(5))
// res1: Boolean = false
```

lxol:
without my change the snippet always returns false

josef-vlach:
What's the case when you need 
```
lr.contains(NumberResult(2))
```
to return true?

`contains` is linked to choice / revealing choice which always yields `OptionResult`  not a `NumberResult`.

By this proposed change you are allowing weird expressions to exists like `myChoice contains mySalary`, this doesn't make much sense and we should not rely on them

lxol:
`=` is the way to check if  ListResult contains the value. That functionality works already.

PR #1518, opened 2022-06-09
[GFORMS-1780] Update file component to improve user interaction

lxol:


public/javascripts/gformFileUpload.js
```diff
@@ -5,543 +5,61 @@
...
-        return false;
-      }
-      disableSubmitButton();
-      $(".govuk-error-message").remove();
-      var $input = $(e.currentTarget);
-      var dataset = $input[0].dataset;
-      var formTemplateId = dataset.formTemplateId;
-      var accessCode = dataset.accessCode;
-      var fileId = dataset.fileId;
-      var sectionNumber = dataset.sectionNumber;
```
josef-vlach, line 249: It seems like these attributes can be deleted https://github.com/hmrc/gform-frontend/blob/lxol/GFORMS-1780/app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala#L1099-L1105

lxol: ok, looks like dataset is not being used anywhere else

josef-vlach (approved):

PR #1526, opened 2022-06-17
GFORMS-1761 - Fix rounding issue when 2 rounded top level expressions…

DursunGundogan:
… are added together

### DIFF
--- a/app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
+++ b/app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
@@ -238,7 +238,12 @@ case class EvaluationResults(
   def evalTyped(er: ExpressionResult, tpe: ExplicitExprType): ExpressionResult =
     tpe match {
       case ExplicitExprType.Sterling(roundingMode) =>
-        er.withNumberResult(bigDecimal => NumberSetScale.setScale(bigDecimal, 2, roundingMode))
+        er match {
+          case NumberResult(_) =>
+            er.withNumberResult(bigDecimal => NumberSetScale.setScale(bigDecimal, 2, roundingMode))
+          case _ => er
+        }
+
       case _ => er
     }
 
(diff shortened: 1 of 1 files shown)


app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -238,7 +238,12 @@ case class EvaluationResults(
   def evalTyped(er: ExpressionResult, tpe: ExplicitExprType): ExpressionResult =
     tpe match {
       case ExplicitExprType.Sterling(roundingMode) =>
-        er.withNumberResult(bigDecimal => NumberSetScale.setScale(bigDecimal, 2, roundingMode))
+        er match {
+          case NumberResult(_) =>
+            er.withNumberResult(bigDecimal => NumberSetScale.setScale(bigDecimal, 2, roundingMode))
+          case _ => er
+        }
```
josef-vlach, line 245: How is this not the same as 
```
er.withNumberResult(bigDecimal => NumberSetScale.setScale(bigDecimal, 2, roundingMode))
```
? 

Can you provide minimal example where this makes a difference?

DursunGundogan: I think we should round only `NumberResult`, not `ListResult`

DursunGundogan: Reverted

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -203,8 +203,15 @@ case class EvaluationResults(
...
-          case _               => unsupportedOperation("Number")(expr)
+          case lrs: ListResult =>
+            lrs.list.fold(NumberResult(0)) { case (a, b) =>
+              val roundedB = field1 match {
+                case Typed(_, tpe) => evalTyped(b, tpe)
+                case _             => b
+              }
+              a + roundedB
+            }
+          case _ => unsupportedOperation("Number")(expr)
```
josef-vlach, line 214: This is weird. Why is rounding logic living in the branch of code where we are adding numbers together? This is not systematic solution to the problem.

PR #1546, opened 2022-07-26
GFORMS-1835: Add boolean expression capability for country lookup column values

lxol:


app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -267,13 +269,16 @@ case class EvaluationResults(
     StringResult(s"$monthAsText $year")
   }
 
+  import org.slf4j.LoggerFactory
+  private val logger = LoggerFactory.getLogger(getClass)
```
josef-vlach, line 273: Remove this

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -267,13 +269,16 @@ case class EvaluationResults(
...
   private def evalString(
     typeInfo: TypeInfo,
     recData: RecData[SourceOrigin.OutOfDate],
     booleanExprResolver: BooleanExprResolver,
     evaluationContext: EvaluationContext
   ): ExpressionResult = {
-
+    logger.error(s"[evalString] typeInfo $typeInfo")
+    logger.error(s"[evalString] recData $recData")
```
josef-vlach, line 281: Remove this

app/uk/gov/hmrc/gform/lookup/LookupOptions.scala
```diff
@@ -49,8 +49,11 @@ object LookupOptions {
   def getLookupValue(lookupInfo: LookupInfo, columnName: String): Option[String] =
     (lookupInfo, columnName) match {
       // format: off
-      case (CountryLookupInfo(_, _, _, _, region), CsvColumnName.region)                  => Some(region.region)
-      case (CountryLookupInfo(id, _, _, _, _), CsvColumnName.countryCode)                 => Some(id.id)
+      case (CountryLookupInfo(_, _, _, _, region, _, _, _), CsvColumnName.region)               => Some(region.region)
+      case (CountryLookupInfo(id, _, _, _, _, _, _, _), CsvColumnName.countryCode)              => Some(id.id)
+      case (CountryLookupInfo(_, _, _, _, _, inEU, _, _), CsvColumnName.inEU)                   => Some(inEU.inEU)
+      case (CountryLookupInfo(_, _, _, _, _, _, inEEA, _), CsvColumnName.inEEA)                   => Some(inEEA.inEEA)
+      case (CountryLookupInfo(_, _, _, _, _, _, _, inEFTA), CsvColumnName.inEFTA)                   => Some(inEFTA.inEFTA)
```
josef-vlach, line 56: Please format this manually

app/uk/gov/hmrc/gform/LookupLoader.scala
```diff
@@ -102,19 +108,42 @@ class LookupLoader {
     keywords: String,
     priority: String,
     region: String,
+    inEU: String,
+    inEEA: String,
+    inEFTA: String,
```
josef-vlach, line 113: I've thought this would be `Map[String, String]` and it will contain all columns from cvs. Let'a hope we won't be adding these new columns every week 🤞

josef-vlach (changes requested):
It doesn't compile

```
not found: type LookupInEFTA
not found: type LookupInEEA
```

lxol:
> 



> It doesn't compile
> 
> ```
> not found: type LookupInEFTA
> not found: type LookupInEEA
> ```

oops, forgot to add 2 new files

app/uk/gov/hmrc/gform/LookupLoader.scala
```diff
@@ -102,19 +108,42 @@ class LookupLoader {
...
+      LookupRegion,
+      LookupInEU,
+      LookupInEEA,
+      LookupInEFTA
+    )
 
     val headerDecoder: HeaderDecoder[ColumnData] =
-      HeaderDecoder.decoder(englishLabel, welshLabel, id, keywords, priority, region)(
-        (_: LookupLabel, _: LookupLabel, _: LookupId, _: LookupKeywords, _: LookupPriority, _: LookupRegion)
+      HeaderDecoder.decoder(englishLabel, welshLabel, id, keywords, priority, region, inEU, inEEA, inEFTA)(
```
josef-vlach, line 130: Sorry for late notice, but I don't think this is right.
We should not read `inEU`, `inEEA` and `inEFTA` explicitly. We should just read whole csv row into `Map[String, String]` (where keys are column names) and base a lookup of column value from `CsvCountryCheck ` on this Map. This will give us nice point of failure. Currently if `column` doesn't exists in csv it doesn't crash

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/LookupLoader.scala
lxol: this method (`readCountries`) produces `CountryLookupInfo` with structure that replicates the csv column and is used for Ajax lookups. It is also used in a lot of places in `match` expressions.  I think if we can't easy make it based on the `Map`.
Shall we create a separate `CountryLookupCheckInfo` for `CsvCountryCheck` only ? It will be based on `Map`.

josef-vlach: I'would prefer to extend `CountryLookupInfo` to look like this:
```
final case class CountryLookupInfo(
  id: LookupId,
  index: Int,
  keywords: LookupKeywords,
  priority: LookupPriority,
  region: LookupRegion,
  columns: Map[String, String]
) extends LookupInfo
```
I feel that to have one object with entirety data from csv row is a good thing.
I think it should be possible to read into the Map at this point: https://github.com/hmrc/gform-frontend/blob/main/app/uk/gov/hmrc/gform/LookupLoader.scala#L117
(We have `index: Int` which I suppose is row number, so we can read that row)

lxol: ok, will to it this way, :bow:

app/uk/gov/hmrc/gform/lookup/LookupOptions.scala
```diff
@@ -49,8 +49,9 @@ object LookupOptions {
   def getLookupValue(lookupInfo: LookupInfo, columnName: String): Option[String] =
     (lookupInfo, columnName) match {
       // format: off
-      case (CountryLookupInfo(_, _, _, _, region), CsvColumnName.region)                  => Some(region.region)
-      case (CountryLookupInfo(id, _, _, _, _), CsvColumnName.countryCode)                 => Some(id.id)
+      case (CountryLookupInfo(_, _, _, _, region, _), CsvColumnName.region)               => Some(region.region)
+      case (CountryLookupInfo(id, _, _, _, _, _), CsvColumnName.countryCode)              => Some(id.id)
+      case (CountryLookupInfo(_, _, _, _, _, columns), column)                            => columns.get(column)
```
josef-vlach, line 54:
```suggestion
      case (CountryLookupInfo(_, _, _, _, _, columns), column)                            => Some(columns.getOrElse(column, throw new Exception(s"Invalid column name $column")))
```

app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
```diff
@@ -591,16 +600,25 @@ case class AuthCacheWithForm(
 case class AuthCacheWithoutForm(
   retrievals: MaterialisedRetrievals,
   formTemplate: FormTemplate,
-  role: Role
+  role: Role,
+  lookupOptions: LocalisedLookupOptions
```
josef-vlach, line 604:
```suggestion
  countryLookupOptions: LocalisedLookupOptions
```
And remove line 606

app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
```diff
@@ -546,15 +551,18 @@ sealed trait AuthCache {
...
 }
 
 case class AuthCacheWithForm(
   retrievals: MaterialisedRetrievals,
   form: Form,
   formTemplateWithRedirects: FormTemplateWithRedirects,
   role: Role,
-  accessCode: Option[AccessCode]
+  accessCode: Option[AccessCode],
+  lookupOptions: LocalisedLookupOptions
```
josef-vlach, line 563:
```suggestion
  countryLookupOptions: LocalisedLookupOptions
```
And remove line 565

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/lookup/LookupOptions.scala
lxol: it will change the current behaviour when we just ignore an non-existent column. We even have a test for that. Is that okay?
https://github.com/hmrc/gform-frontend/blob/2c2856d47f0270b31541871b4c2b6dec04627b30/test/uk/gov/hmrc/gform/lookup/LookupOptionsSpec.scala#L88

josef-vlach (approved):

PR #1548, opened 2022-07-27
GFORMS-1830 - Fix upscan issue where data entered in fields lost when…

DursunGundogan:
… file is uploaded on same page

### DIFF
--- a/public/javascripts/gformFileUpload.js
+++ b/public/javascripts/gformFileUpload.js
@@ -27,7 +27,6 @@
     var submitButton = form.find(".govuk-button--secondary")
     var formComponentId = input.attr("id");
     var uploadedFiles = $("#" + formComponentId + "-files")
-    var frm = $("#gf-form");
     submitButton.css("display", "")
     submitButton.on("click", function(evt) {
       formGroup.hide()
@@ -35,14 +34,25 @@
       submitButton.css("display", "none")
       evt.preventDefault();
 
+      var gfForm = $("#gf-form");
       $.ajax({
-          type: frm.attr('method'),
-          url: frm.attr('action'),
-          data: frm.serialize()
-      }).then(function (){
-          submitButton.submit();
+          type: gfForm.attr("method"),
+          url: gfForm.attr("action"),
+          data: gfForm.serialize()
+      }).done(function (){
+          var upscanForm = $("#gf-upscan-"+formComponentId);
+          var data = new FormData(upscanForm[0]);
+
+          $.ajax({
+              type: upscanForm.attr("method"),
+              enctype: "multipart/form-data",
+              url: upscanForm.attr("action"),
+              data: data,
+              processData: false,
+              contentType: false,
+              cache: false
+          });
       });
-      return false;
     });
   }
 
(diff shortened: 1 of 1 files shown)


public/javascripts/gformFileUpload.js
```diff
@@ -40,7 +40,7 @@
           url: frm.attr('action'),
           data: frm.serialize()
       }).then(function (){
-          frm.submit();
+          submitButton.submit();
```
josef-vlach, line 43: Testing this locally it seems not to work. Are you confident this is correct solution?

public/javascripts/gformFileUpload.js
```diff
@@ -27,22 +27,32 @@
...
-      }).then(function (){
-          frm.submit();
+          type: gfForm.attr("method"),
+          url: gfForm.attr("action"),
+          data: gfForm.serialize()
+      }).done(function (){
+          var upscanForm = $("#gf-upscan-"+formComponentId);
+          var data = new FormData(upscanForm[0]);
+
+          $.ajax({
```
josef-vlach, line 46: Hm, let's not do this second ajax call. This is getting complicated again. 

I think we should just click on `submitButton` programmatically (we need to change onClick callback prior to that obviously, as not to end up in a loop)

public/javascripts/gformFileUpload.js
```diff
@@ -27,22 +27,32 @@
...
-          type: frm.attr('method'),
-          url: frm.attr('action'),
-          data: frm.serialize()
-      }).then(function (){
-          frm.submit();
+          type: gfForm.attr("method"),
+          url: gfForm.attr("action"),
+          data: gfForm.serialize()
+      }).done(function (){
+          var upscanForm = $("#gf-upscan-"+formComponentId);
```
josef-vlach, line 43: This is `upscan` specific, so we cannot do that. but it doesn't matter since this will go away anyway, see my previous comments.

josef-vlach (changes requested):

josef-vlach:
Here is suggestion:

```
   var fileSubmit = function(form, button) {
     var formGroup = form.find(".govuk-form-group");
     var input = formGroup.find(".govuk-file-upload");
     var formComponentId = input.attr("id");
     var uploadedFiles = $("#" + formComponentId + "-files");

     button.css("display", "none");
     formGroup.hide();
     uploadedFiles.empty().append(startProgressBar());

     return true;
   }

   var dataSubmit = function(form, dataForm, button) {
     $.ajax({
       type: dataForm.attr('method'),
       url: dataForm.attr('action'),
       data: dataForm.serialize()
     }).then(function (){
       button.on("click", function(e) {
         fileSubmit(form, button);
       });
       button.click();
     });
   }

  function handleFileUpload(e) {
    var form = $(e.target).closest("form")
    var submitButton = form.find(".govuk-button--secondary")
    var dataForm = $("#gf-form");
    submitButton.css("display", "")
    submitButton.on("click", function(e) {
      dataSubmit(form, dataForm, submitButton)
    });
  }

```

josef-vlach (approved):

PR #1593, opened 2022-09-01
GFORMS-1890: Add ability to retrieve name, email, date of birth and address from auth record

lxol:


app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
```diff
@@ -400,24 +457,32 @@ class AuthenticatedRequestActions(
...
+              cache,
+              recalculation
+            )
 
-        formModelOpticsUpd =
-          formModelOptics
-            .modify(_.formModelVisibilityOptics.recalculationResult.evaluationContext.formTemplateId)
-            .setTo(formUpd.formTemplateId)
+        (cacheUpdated, formModelOpticsUpdated) <-
+          maybeUpdateItmpCacheAndOptics(
```
josef-vlach, line 467: This is executed on every request. As discussed before, we should do it only once when user enters the form and so it should be implemented in `NewFormController.scala`

josef-vlach (changes requested):

josef-vlach (approved):

PR #1594, opened 2022-09-01
GFORMS-1452 - Implement Company Registration Number retrieval within …

DursunGundogan:
…form

### DIFF
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -136,6 +136,12 @@ microservice {
         host = localhost
         port = 9871
       }
+      company-information {
+         protocol = https
+         host = api.company-information.service.gov.uk
+         port = 443
+         apiKey = 20757a9d-c286-4fbc-bd7f-3c822bac9517
+      }
     }
 }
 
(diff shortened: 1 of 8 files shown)


conf/application.conf
```diff
@@ -136,6 +136,12 @@ microservice {
         host = localhost
         port = 9871
       }
+      company-information {
+         protocol = https
+         host = api.company-information.service.gov.uk
+         port = 443
+         apiKey = 20757a9d-c286-4fbc-bd7f-3c822bac9517
```
josef-vlach, line 143: Isn't this supposed to be secret? Whose key is this?

josef-vlach (commented):
LGTM, except that `apiKey` in plain text

conf/application.conf
DursunGundogan: Right, it needs to be encrypted. It's a developer key

VlachJosef (approved):

josef-vlach (approved):

PR #1597, opened 2022-09-07
GFORMS-1888 - Add a feature to allow a choice to redirect a user to a…

DursunGundogan:


app/uk/gov/hmrc/gform/gform/FormController.scala
```diff
@@ -550,7 +550,9 @@ class FormController(
                           )
                       }
 
-                    updatePostcodeLookup.fold(continueJourney) { case (formComponentId, _) =>
+                    val continueJourneyWithoutRedirect = maybeRedirect.fold(continueJourney)(r => Redirect(r))
```
josef-vlach, line 553: `continueJourneyWithoutRedirect` deserves better name, since you just did the redirect, so the name is misleading

josef-vlach (approved):

josef-vlach (approved):

PR #1622, opened 2022-09-28
GFORMS-1452 - Implement Company Registration Number retrieval within …

DursunGundogan:
…form


app/uk/gov/hmrc/gform/models/CompanyProfile.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  object RegisteredAddress {
+    private val templateReads: Reads[RegisteredAddress] =
+      ((__ \ "address_line_1").read[String] and
+        (__ \ "address_line_2").readNullable[String] and
+        (__ \ "postal_code").read[String] and
+        (__ \ "locality").read[String] and
+        (__ \ "region").readNullable[String])(RegisteredAddress.apply _)
+
+    implicit val format: OFormat[RegisteredAddress] = OFormatWithTemplateReadFallback(templateReads)
```
josef-vlach, line 44: Wow, we should not have this `OFormatWithTemplateReadFallback` on the frontend at all.

app/uk/gov/hmrc/gform/models/CompanyProfile.scala
```diff
@@ -0,0 +1,62 @@
...
+    region: Option[String]
+  )
+
+  object RegisteredAddress {
+    private val templateReads: Reads[RegisteredAddress] =
+      ((__ \ "address_line_1").read[String] and
+        (__ \ "address_line_2").readNullable[String] and
+        (__ \ "postal_code").read[String] and
+        (__ \ "locality").read[String] and
+        (__ \ "region").readNullable[String])(RegisteredAddress.apply _)
```
josef-vlach, line 42: Can we derive this `Reads` and pass it from the backend, instead of sending raw response?

app/uk/gov/hmrc/gform/models/CompanyProfile.scala
```diff
@@ -0,0 +1,62 @@
...
+    status: String,
+    registeredAddress: RegisteredAddress
+  )
+
+  object Response {
+
+    private val templateReads: Reads[Response] =
+      ((__ \ "company_name").read[String] and
+        (__ \ "company_status").read[String] and
+        (__ \ "registered_office_address").read[RegisteredAddress])(Response.apply _)
```
josef-vlach, line 58: Can we derive this Reads and pass it from the backend, instead of sending raw response?

josef-vlach (changes requested):

PR #1650, opened 2022-11-08
GFORMS-1966 - Change behaviour of back link

DursunGundogan:


app/uk/gov/hmrc/gform/gform/FormController.scala
```diff
@@ -326,17 +326,25 @@ class FormController(
...
-                sectionTitle4Ga,
-                SuppressErrors.Yes,
-                ff
-              )
-          ).pure[Future]
+          ff match {
+            case FastForward.CYA(_, _) =>
+              Redirect(
+                routes.SummaryController
+                  .summaryById(cache.formTemplateId, maybeAccessCode, sectionNumber.toCoordinates, None)
```
lxol, line 333: I think we should got to this route only for 
```
case FastFrorward.CYA(_, None)
```
If the second argument is `Some(sectionNumber)` it means that it is either ATL CYA pager or mini summary list CYA page

lxol: I think we should go to this `sectionNumber` instead of the final CYA page

app/uk/gov/hmrc/gform/gform/FormController.scala
```diff
@@ -654,61 +662,57 @@ class FormController(
...
+                      fastForward
+                    )
+
+                val fastForwardSn: Call = {
+                  fastForward match {
+                    case FastForward.CYA(_, None) =>
+                      routes.SummaryController
+                        .summaryById(cache.formTemplateId, maybeAccessCode, commingFromSn.toCoordinates, None)
+                    case FastForward.CYA(_, Some(to)) =>
+                      backRoute(to, FastForward.Yes)
```
lxol, line 685: if we use FastForward.Yes here we won't be able to go back several times and then got forward step by step

public/javascripts/gformFormActionHandlers.js
```diff
@@ -1,87 +0,0 @@
...
-          $hint.attr('id', id).attr('aria-live', 'polite');
-          var $textarea = $hint.siblings('textarea');
-          var existingAttr = $textarea.attr('aria-describedby')
-            if(existingAttr){
-                $textarea.attr('aria-describedby', existingAttr + " " + id);
-            } else {
-                $textarea.attr('aria-describedby', id);
-            }
-
-          })
```
josef-vlach, line 72: Can this really be removed? Be aware that this file apparently handles more stuff than back link click only.

josef-vlach (changes requested):

josef-vlach (approved):

PR #1658, opened 2022-11-15
GFORMS-1966 - Change behaviour of back link

DursunGundogan:


app/uk/gov/hmrc/gform/gform/FormController.scala
```diff
@@ -448,7 +448,7 @@ class FormController(
             // TODO: Check why FormSummary and TaskSummary ignored
             ff match {
               case FastForward.CYA(from, to) =>
-                (sectionNumber, FastForward.CYA(from, SectionOrSummary.Section(sectionNumber.increment)))
+                (sectionNumber, FastForward.CYA(sectionNumber, SectionOrSummary.Section(from)))
```
lxol, line 451: there is a regression in this change. When we click change on the addToList summary page we should go through all pages in ATL one by one. With this change it just jumps back to the ATL summary page

DursunGundogan: I think it does not match CYA(from, to) pattern in this point. so back button takes me to `to` section, not `from` section

PR #1664, opened 2022-11-18
GFORMS-1930: After changing an item from the summary / CYA page the user is taken back to task list

lxol:
The original idea that we can use `maybeCooridinates` as `None` to indicate that we need to go to the FormSummary was wrong.  In some cases we still need both coordinates and the flag that indicates that FormSummary was reached.
The fix is to introduce `reachedFormSummary` flag, leave the current logic intact for `reachedFormSummary = false` and for other   case have the following logic

| hasVisibleTask List Summary | isFormValid | redirect to |
|--------------------------|--------------|------------------------------------|
| no | no | landing page|
| no | yes | Form page|
| yes | no | landing page|
| yes | yes| Form page|


josef-vlach (changes requested):
Please change commit message to include whole Jira ticket description.

Saying `bug fix` is not much helpful to quickly see what the commit is about.

josef-vlach (approved):

PR #1680, opened 2022-12-01
GFORMS-1989 - Deprecate limit property in addToList and replace with …

DursunGundogan:
…repeatsUntil and repeatsWhile


app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -316,32 +328,13 @@ class SectionRenderingService(
...
-                      htmlForInformationMessage(info, infoType, infoText),
-                      html.form.snippets.hidden(formComponent.id.value, "1")
-                    )
-                  )
-                )
-              } else {
-                AddToListLimitReached.No
-              }
-            }
+    val evalRepeatsUntil = bracket.source.repeatsUntil
```
josef-vlach, line 331: Any usage of `bracket.source` like `bracket.source.repeatsUntil` seems like a bug. `repeatsUntil` includeIf should be able to refer to fields in current iteration etc.

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -359,7 +352,28 @@ class SectionRenderingService(
...
+      radios.copy(items = items.copy(head = items.head.copy(checked = true)).toList, fieldset = hiddenFieldset)
+
+    lazy val addAnotherQuestionWithYes: Html =
+      new components.GovukRadios(govukErrorMessage, govukFieldset, govukHint, govukLabel)(radiosWithYes)
+
+    lazy val radiosWithNo =
+      radios.copy(items = items.copy(tail = items.tail.map(_.copy(checked = true))).toList, fieldset = hiddenFieldset)
+
+    lazy val addAnotherQuestionWithNo: Html =
+      new components.GovukRadios(govukErrorMessage, govukFieldset, govukHint, govukLabel)(radiosWithNo)
```
josef-vlach, line 365: Why `lazy` vals?

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -359,7 +352,28 @@ class SectionRenderingService(
...
+
+    val addAnotherQuestionSnippets =
+      (bracket.source.repeatsWhile, bracket.source.repeatsUntil) match {
+        case (Some(_), Some(_)) =>
+          if (evalRepeatsWhile || evalRepeatsUntil) addAnotherQuestionWithNo else addAnotherQuestion
+        case (Some(_), None) => if (evalRepeatsWhile) addAnotherQuestionWithYes else addAnotherQuestion
+        case (None, Some(_)) => if (evalRepeatsUntil) addAnotherQuestionWithNo else addAnotherQuestion
+        case _               => addAnotherQuestion
+      }
```
josef-vlach, line 374: This doesn't make sense. If I understand this correctly it should be simply:

```suggestion
      if (!evalRepeatsWhile)
        addAnotherQuestionWithNo
      else if (!evalRepeatsUntil)
        addAnotherQuestionWithYes
      else addAnotherQuestion
```

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
DursunGundogan: I think we should check that **repeatsWhile** and **repeatsUntil** has an includeIf before comparison. Because if any of them does not exist, it returns `false` by default. It seems your suggestion does not work for **repeatsWhile** condition.

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
DursunGundogan: Removed.

josef-vlach (approved):

PR #1694, opened 2022-12-20
GFORMS-1997: Make navigation consistent following updates to fast forward and back link behaviour

lxol:


app/uk/gov/hmrc/gform/gform/handlers/FormValidator.scala
```diff
@@ -91,24 +90,16 @@ class FormValidator(implicit ec: ExecutionContext) {
     validationResult.toFormValidationOutcome(enteredVariadicFormData)
   }
 
-  def fastForwardValidate(
+  def maybeGetInvalidSectionNumber(
```
josef-vlach, line 93: Why is this called `maybeGetInvalidSectionNumber`? Why e need invalid section number?

app/uk/gov/hmrc/gform/gform/FormController.scala
```diff
@@ -862,4 +912,35 @@ class FormController(
...
+    formModel: FormModel[Visibility]
+  ): List[FastForward] = {
+    val sectionNumber = formModel.visibleSectionNumber(browserSectionNumber)
+    val ff = rawFastForward
+      .filterNot {
+        case FastForward.CYA(SectionOrSummary.Section(sn)) => sectionNumber >= sn
+        case FastForward.StopAt(sn)                        => sectionNumber >= sn
+        case _                                             => false
+      }
+    removeDuplications(ff)
```
josef-vlach, line 928: What is a purpose of this? Where does the duplicates comes from?

app/uk/gov/hmrc/gform/gform/handlers/FormValidator.scala
lxol: I think this is `diff` shows the change as renaming. In fact I just created a new method `maybeGetInvalidSectionNumber` that checks if there is an invalid section number in range. `fastForwardValidate` is still there

app/uk/gov/hmrc/gform/gform/FormController.scala
lxol: I wanted to separate processes of adding the required `FastForward` to the `List[FastForward]` from the process of removing. So when I add the new `FastForward` I am not concern with what is in the list already - just add a required at this particular point to the list.
The removal  happens at the beginning of processing of 4 endpoints that takes `List[FastForward]`. It does the following:
 - removes duplications
 - removes FastForward with destination section number higher than the current section number. (

app/uk/gov/hmrc/gform/gform/handlers/FormValidator.scala
josef-vlach: Still I have sort of problem with this name. 

Looking at the implementation it is not only about invalid section, is it? 
If that is right, can we rename this method, perhaps it can hint what the result of this method is intended to be used for?

lxol: how about `def mustBeVisitedSectionNumber` ?

josef-vlach (approved):

PR #1699, opened 2022-12-28
GFORMS-2013: Update 15 minute timeout counter periodically when user …

lxol:
…is active on the keyboard

### DIFF
--- a/public/javascripts/gformSessionTimeout.js
+++ b/public/javascripts/gformSessionTimeout.js
@@ -1,18 +1,18 @@
 (function(global) {
   "use strict";
 
-  var $ = global.jQuery;
-  var GOVUK = global.GOVUK || {};
-  var lang = (global.gform && global.gform.lang) || "en";
-
   function GformSessionTimeout() {
     var self = this;
 
+    var enableBroadcastTimestamp = true;
     function init() {
       $("input").keypress(broadcastTimestamp);
     }
 
     function broadcastTimestamp() {
+      if (!enableBroadcastTimestamp) return;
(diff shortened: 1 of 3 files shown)


public/javascripts/gformSessionTimeout.js
```diff
@@ -0,0 +1,36 @@
...
+
+  var $ = global.jQuery;
+  var GOVUK = global.GOVUK || {};
+  var lang = (global.gform && global.gform.lang) || "en";
+
+  function GformSessionTimeout() {
+    var self = this;
+
+    function init() {
+      $("input").keypress(broadcastTimestamp);
```
josef-vlach, line 12: Can we throttle this and broadcast only every 5 minutes or so?

public/javascripts/gformSessionTimeout.js
```diff
@@ -0,0 +1,36 @@
+(function(global) {
+  "use strict";
+
+  var $ = global.jQuery;
+  var GOVUK = global.GOVUK || {};
+  var lang = (global.gform && global.gform.lang) || "en";
```
josef-vlach, line 6: Not needed

josef-vlach (changes requested):

public/javascripts/gformSessionTimeout.js
lxol: done

public/javascripts/gformSessionTimeout.js
lxol: removed

josef-vlach (approved):

PR #1704, opened 2023-01-05
GFORMS-1997: fix a double click bug

lxol:
Please see Navigation.md
NavFLOW-021: the wrong behaviour - step 3 goes to Page1

### DIFF
--- a/app/uk/gov/hmrc/gform/gform/handlers/FormValidator.scala
+++ b/app/uk/gov/hmrc/gform/gform/handlers/FormValidator.scala
@@ -156,6 +156,7 @@ class FormValidator(implicit ec: ExecutionContext) {
     fastForward: List[FastForward],
     maybeSectionNumber: Option[SectionNumber]
   ): Future[SectionOrSummary] = {
+
     val maybeCoordinates = maybeSectionNumber.flatMap(_.toCoordinates)
     val formModelOptics: FormModelOptics[DataOrigin.Browser] = processData.formModelOptics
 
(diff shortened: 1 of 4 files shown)


app/uk/gov/hmrc/gform/models/FormModel.scala
```diff
@@ -302,7 +302,7 @@ case class FormModel[A <: PageMode](
       // User is trying to see invisible page, so we need to send him to appropriate SectionNumber instead
       availableSectionNumbers match {
         case Nil       => throw new IllegalArgumentException(s"Cannot find valid sectionNumber for $sectionNumber.")
-        case head :: _ => availableSectionNumbers.reverse.find(_ < sectionNumber).getOrElse(head)
+        case head :: _ => availableSectionNumbers.find(_ > sectionNumber).getOrElse(head)
```
josef-vlach, line 305: This will affect how navigation to invisible page is handled. Currently it goes to previous visible page (if one exists).
With this change it will go to next (possible never visited section).
Also with this change when you visit non existent section number bigger than last section number it navigates to first page instead of last visible page (as I'd expect).

There must be a better fix.

Or else we need to formalise how navigation to invisible pages should be handled.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/gform/handlers/FormValidator.scala
```diff
@@ -191,13 +191,15 @@ class FormValidator(implicit ec: ExecutionContext) {
           }
         )
       case FastForward.StopAt(to) :: xs =>
+        val maxTo = List(Some(to), nextFrom).max.get
```
josef-vlach, line 194: Please let's not use unsafe `.get` in production code

```suggestion
        val maxTo = nextFrom.fold(to)(n => List(to, n).max)
```

lxol: committed your changes...

josef-vlach (approved):

PR #1728, opened 2023-01-19
GFORMS-1991: Allow for ATL data to be presented as bulletedList or nu…

lxol:
…mberedList in info fields

### DIFF
--- a/app/uk/gov/hmrc/gform/gform/AcknowledgementController.scala
+++ b/app/uk/gov/hmrc/gform/gform/AcknowledgementController.scala
@@ -172,7 +172,8 @@ class AcknowledgementController(
               SummaryPagePurpose.ForUser,
               Some(summarySectionDeclaration),
               None,
-              maybeTabularFormat
+              maybeTabularFormat,
+              Some(cache.formTemplate.formName.value)
             )
         pdfSource <- pdfService.generatePDF(pdfHtml)
       } yield Result(
--- a/app/uk/gov/hmrc/gform/pdf/PDFRenderService.scala
+++ b/app/uk/gov/hmrc/gform/pdf/PDFRenderService.scala
@@ -51,7 +51,8 @@ class PDFRenderService(
     purpose: SummaryPagePurpose,
     summaryDeclaration: Option[Html],
     maybeDraftText: Option[String] = None,
-    maybeTabularFormat: Option[Boolean] = None
+    maybeTabularFormat: Option[Boolean] = None,
+    maybeFormName: Option[String] = None
   )(implicit
     request: Request[_],
     messages: Messages,
@@ -91,7 +92,9 @@ class PDFRenderService(
             title,
             pdfModel,
             maybeHeaderFooter,
-            cache.formTemplate
+            cache.formTemplate,
+            maybeFormName,
+            maybeDraftText
           ).toString
       }
       PdfHtml(html)
--- a/app/uk/gov/hmrc/gform/views/summary/summaryTabularPdf.scala.html
+++ b/app/uk/gov/hmrc/gform/views/summary/summaryTabularPdf.scala.html
@@ -25,7 +25,9 @@
 title: String,
 summaryDatas: List[SummaryData],
 maybeHeaderFooter: Option[HeaderFooter],
-formTemplate: FormTemplate)(implicit request: Request[_], messages: Messages, l:LangADT, sse: SmartStringEvaluator)
+formTemplate: FormTemplate,
+maybeFormName: Option[String],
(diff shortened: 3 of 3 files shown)


app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -330,6 +344,32 @@ case class EvaluationResults(
...
+          addressAtoms.map(atom => recData.variadicFormData.get(atom))
+        val addressLines = variadicValues.collect { case Some(VariadicValue.One(value)) if value.nonEmpty => value }
+        addressLines.mkString(", ")
+      }
+    def atlValues(fcId: FormComponentId): List[String] =
+      recData.variadicFormData
+        .forBaseComponentIdLessThen(fcId.modelComponentId)
+        .map { case (id, _) =>
+          val isHidden = exprMap.get(FormCtx(id.toFormComponentId)).fold(false)(_ === Hidden)
+          lxol.pp.log(isHidden)
```
DursunGundogan, line 365: Please remove this log. It causes a compilation error.

lxol: oops

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -508,6 +547,23 @@ case class EvaluationResults(
...
+        val list = atlValues(fcId)
+        StringResult(govukList(list, false))
+      case BulletedList(fcId)
+          if evaluationContext.addressLookup(fcId.baseComponentId) || evaluationContext
+            .overseasAddressLookup(fcId.baseComponentId) =>
+        val list = atlAddresses(fcId)
+        StringResult(govukList(list, true))
+
+      case BulletedList(fcId) =>
+        val list = atlValues(fcId)
```
josef-vlach, line 565: Replace `atlValues(fcId)` with `loop(FormCtx(fcId))` and process resulting `ListResult` accordingly.

Note: `def atlValues` should be eliminated

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -508,6 +547,23 @@ case class EvaluationResults(
...
+            .overseasAddressLookup(fcId.baseComponentId) =>
+        val list = atlAddresses(fcId)
+        StringResult(govukList(list, false))
+      case NumberedList(fcId) =>
+        val list = atlValues(fcId)
+        StringResult(govukList(list, false))
+      case BulletedList(fcId)
+          if evaluationContext.addressLookup(fcId.baseComponentId) || evaluationContext
+            .overseasAddressLookup(fcId.baseComponentId) =>
+        val list = atlAddresses(fcId)
```
josef-vlach, line 561: This special handling of addresses should be eliminated and replaced by `loop(FormCtx(fcId))` which will return `ListResult`.
Unfortunately printing addresses from ATL into smartstrings is not currently implemented, but feel free to fix it.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/commons/MarkDownUtil.scala
```diff
@@ -54,7 +54,7 @@ object MarkDownUtil {
   def markDownParser(ls: LocalisedString)(implicit l: LangADT): Html = markDownParser(ls.value)
 
   def markDownParser(ls: SmartString)(implicit sse: SmartStringEvaluator): Html = {
-    val markDownText = removeSpacesForStrong(ls.valueForMarkdown())
+    val markDownText = removeSpacesForStrong(ls.value())
```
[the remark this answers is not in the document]
lxol: `ls.value`treats smart string as as html. If I leave it `ls.valueForMarkdown` all html will be escaped and will be rendered in a browser literally.

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
lxol: yes, that's why I implemented it separately. I will look into fixing it :+1:

app/uk/gov/hmrc/gform/commons/MarkDownUtil.scala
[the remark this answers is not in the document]
josef-vlach: > will be escaped and will be rendered in a browser literally

That is by design and it needs to keep behaving like that. The reason for that is that user provided input cannot be rendered as html, otherwise users would be able embed javascript into their input which will be executed etc... we don't want that.

PR #1732, opened 2023-01-23
GFORMS-1991: Allow for ATL data to be presented as bulletedList or nu…

lxol:
…mberedList in info fields

fix unescaped user data in the startstrings

### DIFF
--- a/app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
+++ b/app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
@@ -361,22 +361,30 @@ case class EvaluationResults(
       case IfElse(cond, field1: Expr, field2: Expr) =>
         if (booleanExprResolver.resolve(cond)) loop(field1) else loop(field2)
       case Else(field1: Expr, field2: Expr) => loop(field1) orElse loop(field2)
-      case FormCtx(formComponentId: FormComponentId)
+      case expr @ FormCtx(formComponentId: FormComponentId)
           if evaluationContext.addressLookup(formComponentId.baseComponentId) || evaluationContext
             .overseasAddressLookup(formComponentId.baseComponentId) =>
         whenVisible(formComponentId) {
-          val indexedComponentId = formComponentId.modelComponentId.indexedComponentId
-          val addressAtoms: List[ModelComponentId.Atomic] =
-            if (evaluationContext.addressLookup(formComponentId.baseComponentId))
-              Address.fields(indexedComponentId).filter(_.atom != Address.uk)
-            else
-              OverseasAddress.fields(indexedComponentId).toList
+          val modelComponentId = expr.formComponentId.modelComponentId
+          val isModelFormComponentIdPure = modelComponentId.indexedComponentId.isPure
+          val isReferenceIndexed = evaluationContext.indexedComponentIds.exists(
+            _.baseComponentId === modelComponentId.baseComponentId
+          )
+          if (isModelFormComponentIdPure && isReferenceIndexed) {
+            get(expr, fromVariadicValue, evaluationContext)
+          } else {
+            val indexedComponentId = formComponentId.modelComponentId.indexedComponentId
+            val addressAtoms: List[ModelComponentId.Atomic] =
+              if (evaluationContext.addressLookup(formComponentId.baseComponentId))
+                Address.fields(indexedComponentId).filter(_.atom != Address.uk)
+              else
+                OverseasAddress.fields(indexedComponentId).toList
 
-          val variadicValues: List[Option[VariadicValue]] = addressAtoms.map(atom => recData.variadicFormData.get(atom))
-          val addressLines = variadicValues.collect {
-            case Some(VariadicValue.One(value)) if value.nonEmpty => value
+            val variadicValues: List[Option[VariadicValue]] =
+              addressAtoms.map(atom => recData.variadicFormData.get(atom))
+            val addressLines = variadicValues.collect { case Some(VariadicValue.One(value)) if value.nonEmpty => value }
+            ExpressionResult.AddressResult(addressLines)
           }
-          ExpressionResult.AddressResult(addressLines)
         }
       case FormCtx(formComponentId: FormComponentId)
           if evaluationContext.postcodeLookup(formComponentId.baseComponentId) =>
@@ -519,8 +527,8 @@ case class EvaluationResults(
           case _              => unsupportedOperation("String")(expr)
         }
       case RemoveSpaces(fcId) => removeSpaces(fcId, recData)
-      case NumberedList(fcId) => get(FormCtx(fcId), fromVariadicValue, evaluationContext)
-      case BulletedList(fcId) => get(FormCtx(fcId), fromVariadicValue, evaluationContext)
+      case NumberedList(fcId) => loop(FormCtx(fcId))
+      case BulletedList(fcId) => loop(FormCtx(fcId))
     }
 
     loop(typeInfo.expr)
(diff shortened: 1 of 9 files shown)


app/uk/gov/hmrc/gform/gform/ExprUpdater.scala
```diff
@@ -62,6 +62,8 @@ class ExprUpdater(index: Int, baseIds: List[FormComponentId]) {
     case PeriodExt(_, _)                      => expr
     case IndexOf(_, _)                        => expr // This is not explanded on purpose, so it can be used correctly inside ATL
     case RemoveSpaces(formComponentId)        => RemoveSpaces(expandFcId(formComponentId))
+    case NumberedList(formComponentId)        => RemoveSpaces(expandFcId(formComponentId))
+    case BulletedList(formComponentId)        => RemoveSpaces(expandFcId(formComponentId))
```
DursunGundogan, line 66:
```suggestion
    case NumberedList(formComponentId)        => NumberedList(expandFcId(formComponentId))
    case BulletedList(formComponentId)            => BulletedList(expandFcId(formComponentId))
```

app/uk/gov/hmrc/gform/sharedmodel/graph/DependencyGraph.scala
```diff
@@ -185,7 +185,28 @@ object DependencyGraph {
...
+              GraphNode.Expr(FormCtx(fc.id)) ~> GraphNode.Simple(fc.id)
+            )
+          )
+    val govukListNodes = graph.nodes.flatMap { n =>
+      n.value match {
+        case g @ GraphNode.Expr(NumberedList(fcId)) => govukListEdges(g, fcId)
+        case g @ GraphNode.Expr(BulletedList(fcId)) => govukListEdges(g, fcId)
+        case e                                      => List.empty[DiEdge[GraphNode]]
+      }
+    }.toSet
```
josef-vlach, line 207: Why is this here?

josef-vlach (changes requested):
Referring ATL `addressId` outside ATL should work, but it doesn't. First step should be to fix this problem (I've already mentioned it last time) and then build this functionality on top of that solution, otherwise this PR is introducing lots of unnecessary complexity.

app/uk/gov/hmrc/gform/sharedmodel/graph/DependencyGraph.scala
lxol: alternatively we can evaluate `FormCtx(addressid)` to AddressResult(addresslines)` or `ListResult(List(AddressResult))` but we will need to review and update all the usages as they all expect `AddressResult(addresslines)`

josef-vlach: > For `addressId` that refers to ATL address we need `ListResult(AddressResult(...))`

And this is exactly what we need to fix. [This should return]( https://github.com/hmrc/gform-frontend/blob/46f798b67926b82957b1e6ec4aefe5ecf535429a/app/uk/gov/hmrc/gform/eval/EvaluationResults.scala#L355) `ListResult(AddressResult(...))` if `addressId` is ATL component.

app/uk/gov/hmrc/gform/eval/ExpressionResult.scala
```diff
@@ -333,6 +333,13 @@ sealed trait ExpressionResult extends Product with Serializable {
       None
     )(_ => None)(_ => None)(_ => None)
 
+  def govukListRepresentation(typeInfo: TypeInfo, messages: Messages): List[String] =
```
josef-vlach, line 336: Let call this method only `listRepresentation`

app/uk/gov/hmrc/gform/eval/smartstring/SmartStringEvaluatorFactory.scala
```diff
@@ -152,5 +163,29 @@ class RealSmartStringEvaluatorFactory() extends SmartStringEvaluatorFactory {
           .evalAndApplyTypeInfo(typeInfo)
           .optionRepresentation
           .fold(stringRepresentation(typeInfo))(f(_))
+
+      private def govukListRepresentation(typeInfo: TypeInfo, isBulleted: Boolean, markDown: Boolean): String = {
+        formModelVisibilityOptics.evalAndApplyTypeInfo(typeInfo)
```
josef-vlach, line 168: Stray expression

app/uk/gov/hmrc/gform/sharedmodel/VariadicFormData.scala
```diff
@@ -118,6 +119,13 @@ case class VariadicFormData[S <: SourceOrigin](data: Map[ModelComponentId, Varia
       }
     }
 
+  def distinctIndexedComponentIds(modelComponentId: ModelComponentId): List[IndexedComponentId] =
+    forBaseComponentIdLessThen(modelComponentId)
+      .map(_._1.fold(p => p.indexedComponentId)(i => i.indexedComponentId))
```
josef-vlach, line 124:
```suggestion
   forBaseComponentId(modelComponentId.indexedComponentId.baseComponentId)
      .map { case (modelComponentId, _) => modelComponentId.fold(p => p.indexedComponentId)(i => i.indexedComponentId) }
```
`forBaseComponentIdLessThen` should not be used here, we are not in ATL. Also let's get rid of `_._1`

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -343,16 +359,21 @@ case class EvaluationResults(
...
-
-          val variadicValues: List[Option[VariadicValue]] = addressAtoms.map(atom => recData.variadicFormData.get(atom))
-          val addressLines = variadicValues.collect { case Some(VariadicValue.One(value)) if value.nonEmpty => value }
-          ExpressionResult.AddressResult(addressLines)
+          val modelComponentId = formComponentId.modelComponentId
+          val isModelFormComponentIdPure = modelComponentId.indexedComponentId.isPure
+          val isReferenceIndexed = evaluationContext.indexedComponentIds.exists(
+            _.baseComponentId === modelComponentId.baseComponentId
+          )
+          if (isModelFormComponentIdPure && isReferenceIndexed) {
```
josef-vlach, line 367: Now we are duplicating this logic in two places, can we remove the duplication?

josef-vlach (changes requested):
Couple of issues to resolve, see the comments.

But also two observations:

1. if `agentFullName` is used only as `${bulletedList(agentFullName)}` list is not being produced
2. `${addressId}` expression used outside ATL do not produce any output. (`def addressRepresentation(typeInfo: TypeInfo)`  need to handle possibility of a `List`

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
lxol: >. (def addressRepresentation(typeInfo: TypeInfo) need to handle possibility of a List
but it returns `List[String]` where each string is part of the address. If this method takes ListResult(List(AddressResult)) the return type of List[String] wouldn't make sense as it will contain all lines for all addresses in one List
Unless I misunderstood that.

lxol: >if agentFullName is used only as ${bulletedList(agentFullName)} list is not being produced
didn't get that

lxol: > if agentFullName is used only as ${bulletedList(agentFullName)} list is not being produced

do you mean that if `agentFulName` in the smart string without `bulletedList` then it is rendered as empty? I don't know how to render it tbh. Do we have any requirements on how to render list of addresses to a string?

josef-vlach: > but it returns List[String] where each string is part of the address. If this method takes ListResult(List(AddressResult)) the return type of List[String] wouldn't make sense as it will contain all lines for all addresses in one List
Unless I misunderstood that.

If `def addressRepresentation(...): List[String` is called on a `ListResult` of addresses we resulting list should contains individual addresses concatenated by `,<space> `. It is better default than showing nothing.

josef-vlach (approved):

PR #1737, opened 2023-01-31
GFORMS-1991: text constraint bug fix

lxol:


app/uk/gov/hmrc/gform/eval/smartstring/SmartStringEvaluatorFactory.scala
```diff
@@ -165,9 +167,12 @@ class RealSmartStringEvaluatorFactory() extends SmartStringEvaluatorFactory {
           .fold(stringRepresentation(typeInfo))(f(_))
 
       private def govukListRepresentation(typeInfo: TypeInfo, isBulleted: Boolean, markDown: Boolean): String = {
-        val lines = formModelVisibilityOptics
+        val lines0 = formModelVisibilityOptics
```
DursunGundogan, line 170: Do you think we should use a more meaningful name? My idea is like `defaultLines`

DursunGundogan (approved):

PR #1747, opened 2023-02-10
GFORMS-2044: Fields with includeIf are not shown when viewing forms u…

lxol:
…sing specimen link


DursunGundogan (approved):

josef-vlach (changes requested):
This should be implemented on the backend when we are creating a specimen.

PR #1761, opened 2023-02-24
GFORMS-2048 - Add feature to build choices for a choice component bas…

josef-vlach:
…ed on an array of data from ATL

### DIFF
--- a/app/uk/gov/hmrc/gform/eval/smartstring/SmartStringEvaluatorFactory.scala
+++ b/app/uk/gov/hmrc/gform/eval/smartstring/SmartStringEvaluatorFactory.scala
@@ -107,7 +107,7 @@ class RealSmartStringEvaluatorFactory() extends SmartStringEvaluatorFactory {
                       val optionsList = choice.options.zipWithIndex
                         .map {
                           case (OptionData.IndexBased(label, _, _, _), i)        => i.toString -> label
-                          case (OptionData.ValueBased(label, _, _, _, value), i) => value      -> label
+                          case (OptionData.ValueBased(label, _, _, _, value), _) => value      -> label
                         }
                         .toList
                         .toMap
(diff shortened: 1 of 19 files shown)


DursunGundogan (approved):

app/uk/gov/hmrc/gform/eval/smartstring/SmartStringEvaluatorFactory.scala
```diff
@@ -106,8 +106,8 @@ class RealSmartStringEvaluatorFactory() extends SmartStringEvaluatorFactory {
                     case IsChoice(choice) =>
                       val optionsList = choice.options.zipWithIndex
                         .map {
-                          case (OptionData.IndexBased(label, _), i)        => i.toString -> label
-                          case (OptionData.ValueBased(label, value, _), i) => value      -> label
+                          case (OptionData.IndexBased(label, _, _, _), i)        => i.toString -> label
+                          case (OptionData.ValueBased(label, _, _, _, value), i) => value      -> label
```
lxol, line 110: `i` is not used, anywhere it can be `_`
Also in may places we keep adding fields to the case classes and we have to update the matchings.
Wouldn't it be better if we have like
                          case (o: OptionData.ValueBased, _) => o.value -> o.label

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/ComponentType.scala
```diff
@@ -316,6 +316,8 @@ object SummaryDisplayWidth extends Enumeration {
 
 sealed trait OptionData extends Product with Serializable {
   def label: SmartString
+  def hint: Option[SmartString]
+  def dynamic: Option[OptionData.Dynamic]
```
lxol, line 320: shouldn't just:
```
def dynamic: Option[FormComponentId]
```
be enough. We always pattern match OptionData and dynamic is always in the same position.

josef-vlach: `OptionData.Dynamic` is used in quite a few method signatures. I like to see on the first sigh that param is of type  `OptionData.Dynamic` rather than of type `FormComponentId`. 
In the first case reader can immediately guess what that param value is used for without looking into implementation details.

app/uk/gov/hmrc/gform/eval/smartstring/SmartStringEvaluatorFactory.scala
josef-vlach: Using typed based matching would mean to give up majority of guarantees Scala pattern matching is giving us.
I think it is one of the best features to let compiler show you all the places where you may potentially use newly added field (even if you end up not needing it to use in a single place).

PR #1767, opened 2023-03-01
GFORMS-2083: Only retrieve itmp name, email, date of birth and addres…

lxol:
…s from auth record if confidence level is 200 or above


josef-vlach (changes requested):
Please put back constructor based pattern matching

lxol:
> Please put back constructor based pattern matching

done

josef-vlach (approved):

PR #1792, opened 2023-03-20
GFORMS-2099: When retrieving an address from ITMP do not store countr…

lxol:
…y in form data if the address is in the UK

### DIFF
--- a/app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
+++ b/app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
@@ -569,8 +569,20 @@ class AuthenticatedRequestActions(
 
     authorised(predicate)
       .retrieve(itmpRetrievals) { case itmpName ~ itmpDateOfBirth ~ itmpAddress =>
-        ItmpRetrievals(itmpName, itmpDateOfBirth, itmpAddress).pure[Future]
+        val updatedCountry = itmpAddress.flatMap(_.countryName).filterNot(isInUK)
+        val updatedItmpAddress = itmpAddress.map(_.copy(countryName = updatedCountry))
+        ItmpRetrievals(itmpName, itmpDateOfBirth, updatedItmpAddress).pure[Future]
       }
+  }
+
+  def isInUK(country: String): Boolean = country match {
+    case str if str.equalsIgnoreCase("ENGLAND")          => true
+    case str if str.equalsIgnoreCase("SCOTLAND")         => true
+    case str if str.equalsIgnoreCase("WALES")            => true
+    case str if str.equalsIgnoreCase("NORTHERN IRELAND") => true
+    case str if str.equalsIgnoreCase("GREAT BRITAIN")    => true
+    case str if str.equalsIgnoreCase("UNITED KINGDOM")   => true
+    case _                                               => false
   }
 }
 
--- a/app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
+++ b/app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
@@ -575,15 +575,9 @@ class AuthenticatedRequestActions(
       }
   }
 
-  def isInUK(country: String): Boolean = country match {
-    case str if str.equalsIgnoreCase("ENGLAND")          => true
-    case str if str.equalsIgnoreCase("SCOTLAND")         => true
-    case str if str.equalsIgnoreCase("WALES")            => true
-    case str if str.equalsIgnoreCase("NORTHERN IRELAND") => true
-    case str if str.equalsIgnoreCase("GREAT BRITAIN")    => true
-    case str if str.equalsIgnoreCase("UNITED KINGDOM")   => true
-    case _                                               => false
-  }
+  def isInUK(country: String): Boolean = ukParts(country.toUpperCase)
+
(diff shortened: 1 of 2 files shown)


josef-vlach (changes requested):
please can you move this country discarding from json reader, it would be better to do it in business logic.. less coding

lxol:
sorry @josef-vlach  this ticket is not ready yet.

app/uk/gov/hmrc/gform/validation/ValidationUtil.scala
```diff
@@ -121,15 +121,18 @@ object ValidationUtil {
       case IsAddress(Address(_, _, _, Some(AuthCtx(AuthInfo.ItmpAddress)))) =>
```
josef-vlach, line 121: I know this should be flagged in other ticket, but this logic seem incorrect to be here.
Consequence of this being here is that you **can't edit** prepopulate address component. Ability to edit whatever is populated is what we should be able to do.

I would expect prepopulation logic to be included in `SectionRenderingService.scala`

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/validation/ValidationUtil.scala
lxol: yes, that's the problem I haven't thought about. It won't be in SectionRederingService only. It also needs to be rendered properly in  CYA, robotics and pdf as far as I understand.  
Personally I would rather try added a logic to update Addresses that refer to another Address once the original Address is changed. Not sure how easy it would be though.

josef-vlach: > rendered properly in CYA, robotics and pdf as far as I understand.

We should get this for free. Address component rendering should not be tied to the source of data (ie. user entered vs prepopulated from itmpAddress). 

> logic to update Addresses that refer to another Address once the original Address is changed

This sounds completely out of the scope of what prepop should be.

app/uk/gov/hmrc/gform/controllers/AuthenticatedRequestActions.scala
```diff
@@ -569,9 +569,21 @@ class AuthenticatedRequestActions(
...
+
+  def isInUK(country: String): Boolean = country match {
+    case str if str.equalsIgnoreCase("ENGLAND")          => true
+    case str if str.equalsIgnoreCase("SCOTLAND")         => true
+    case str if str.equalsIgnoreCase("WALES")            => true
+    case str if str.equalsIgnoreCase("NORTHERN IRELAND") => true
+    case str if str.equalsIgnoreCase("GREAT BRITAIN")    => true
+    case str if str.equalsIgnoreCase("UNITED KINGDOM")   => true
+    case _                                               => false
+  }
```
josef-vlach, line 586: Can't we keep a `Set` of countries and check that `country` is in that set?

app/uk/gov/hmrc/gform/validation/ValidationUtil.scala
```diff
@@ -121,15 +121,18 @@ object ValidationUtil {
...
           "street1"  -> itmpAddress.flatMap(_.line1).getOrElse(""),
           "street2"  -> itmpAddress.flatMap(_.line2).getOrElse(""),
           "street3"  -> itmpAddress.flatMap(_.line3).getOrElse(""),
           "street4"  -> itmpAddress.flatMap(address => address.line4.map(_ + " ") |+| address.line5).getOrElse(""),
           "postcode" -> itmpAddress.flatMap(_.postCode).getOrElse(""),
-          "country"  -> "",
-          "uk"       -> "true"
+          "uk"       -> itmpAddress.flatMap(_.countryName).filter(_.trim.nonEmpty).fold(true)(_ => false).toString
         )
+        val countryMap =
```
josef-vlach, line 132: This doesn't need to be in extra map. Let's put it this into `atomMap1` (and let's rename `atomMap1` to `atomMap`)

app/uk/gov/hmrc/gform/validation/ValidationUtil.scala
```diff
@@ -165,10 +168,11 @@ object ValidationUtil {
         val linesMap = List("line1", "line2", "line3").zip(lines).toMap
         val otherMap = Map(
           "city"     -> city.headOption.getOrElse(""),
-          "postcode" -> itmpAddress.flatMap(_.postCode).getOrElse(""),
-          "country"  -> itmpAddress.flatMap(_.countryName).getOrElse("")
+          "postcode" -> itmpAddress.flatMap(_.postCode).getOrElse("")
         )
-        val atomMap: Map[String, String] = linesMap ++ otherMap
+        val countryMap =
+          itmpAddress.flatMap(_.countryName).filter(_.trim.nonEmpty).map(c => Map("country" -> c)).getOrElse(Map.empty)
```
josef-vlach, line 174: The same here, this should not be its own map. (Or perhaps I'm missing the reason for this and if so let's put comment into the code to explain it)

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/validation/ValidationUtil.scala
lxol: yes, it's no longer needed. I put "country" back to the original map.

PR #1813, opened 2023-04-06
GFORMS-2113: Add feature to allow emailCodeParameters

lxol:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/EmailParameter.scala
```diff
@@ -54,3 +55,20 @@ object EmailParametersRecalculated {
...
+    val uploadTemplateReads: Reads[EmailCodeParameter] =
+      for {
+        emailTemplateVariable <- (JsPath \ "emailTemplateVariable").read[String]
+        value                 <- (JsPath \ "value").read[LocalisedString]
+      } yield EmailCodeParameter(emailTemplateVariable, value)
+
+    val reads: Reads[EmailCodeParameter] = uploadTemplateReads | mongoFormat
+
+    OFormat(reads, mongoFormat)
+  }
```
josef-vlach, line 73: This should be `derived`, since we are in the frontend, we don't need to deal with reader from raw json template.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/EmailParameter.scala
lxol: done

josef-vlach (approved):

PR #1831, opened 2023-05-03
GFORMS-2163 - Investigate how to host an xlsx and ods file for downlo…

DursunGundogan:
…ad within a form


app/uk/gov/hmrc/gform/gform/DownloadController.scala
```diff
@@ -0,0 +1,50 @@
...
+import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendController
+
+import scala.concurrent.ExecutionContext
+
+class DownloadController(messagesControllerComponents: MessagesControllerComponents, environment: Environment)(implicit
+  ec: ExecutionContext
+) extends FrontendController(messagesControllerComponents) {
+
+  def downloadNipTemplate(fileExt: String): Action[AnyContent] = Action { _ =>
+    val fileName = s"nipClaimScheduleTemplate.$fileExt"
```
lxol, line 31: I think it is potentially dangerous. `fileExt` is a user input. They can provide a path instead of extension and potentially traverse our file system and get access to other files.
Can we sanitize it, like checking that it is in the list of valid extensions?

app/uk/gov/hmrc/gform/gform/DownloadController.scala
```diff
@@ -0,0 +1,82 @@
...
+        if (file.exists()) {
+          Future.successful(
+            Ok.sendFile(
+              content = file,
+              fileName = _ => Some(fileFullName)
+            ).withHeaders(
+              CONTENT_DISPOSITION -> s"inline; filename=$fileFullName",
+              CONTENT_TYPE -> FileInfoConfig.lookup
+                .get(fileExt.value)
+                .getOrElse("application/vnd.openxmlformats-officedocument.spreadsheetml.sheet")
```
josef-vlach, line 72: I think we should not fallback to any particular content type. Can we make this hard error instead?
Also I'm not sure if `FileInfoConfig.lookup` is right to be used here... That lookup is for files we accept as uploads. We should create another lookup for files we want to serve to the users. 

This new lookup could also be mapping from full file name to content type... so we would be able to avoid this `FileExtension` type and all the machinery around it 🤔

josef-vlach (approved):

PR #1836, opened 2023-05-10
GFORMS-2158: Introduce new format positiveWholeSterling

lxol:
…tiveNumber" and "positiveWholeNumber"

### DIFF
--- a/app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
+++ b/app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
@@ -97,6 +97,13 @@ object ComponentValidator {
   val genericNumberErrorMaxdigitPattern                      = "generic.number.error.maxDigit.pattern"
   val genericNumberErrorWholePattern                         = "generic.number.error.whole.pattern"
   val genericNumberErrorPositivePattern                      = "generic.number.error.positive.pattern"
+
+  val genericPositiveWholeSterlingErrorRequired              = "generic.positive.whole.sterling.error.required"
+  val genericPositiveWholeSterlingErrorPattern               = "generic.positive.whole.sterling.error.pattern"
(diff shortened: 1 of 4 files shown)


app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
```diff
@@ -654,6 +671,159 @@ object ComponentValidator {
...
+  private def nonNumericSterlingFailure(
+    fieldValue: FormComponent,
+    value: String
+  )(implicit
+    messages: Messages,
+    sse: SmartStringEvaluator
+  ): ValidatedType[Unit] =
+    (fieldValue.errorShortName, fieldValue.errorShortNameStart, fieldValue.errorExample) match {
+      case (None, None, maybeErrorExample) =>
+        val errorExample = maybeErrorExample.getOrElse(SmartString.blank)
```
josef-vlach, line 703:
```
val errorExample = maybeErrorExample.getOrElse(SmartString.blank)
```
This is duplicated 9 times, can we do something to reduce it to have this logic only once?

app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
```diff
@@ -654,6 +671,159 @@ object ComponentValidator {
...
+  }
+
+  private def nonNumericSterlingFailure(
+    fieldValue: FormComponent,
+    value: String
+  )(implicit
+    messages: Messages,
+    sse: SmartStringEvaluator
+  ): ValidatedType[Unit] =
+    (fieldValue.errorShortName, fieldValue.errorShortNameStart, fieldValue.errorExample) match {
```
josef-vlach, line 701: All matching on `fieldValue.errorExample` are just rebinding the value, rather than doing any matching, can we drop that?

josef-vlach (changes requested):
There seems to be a lot of space for reducing unnecessary duplications in this PR

PR #1845, opened 2023-05-10
GFORMS-2158: GFORMS-2160: Change the error message for the formats "sterling" and "positiveSterling"

lxol:


app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
```diff
@@ -654,6 +673,169 @@ object ComponentValidator {
...
+    }
+
+  private def wholeSterlingFailure(
+    fieldValue: FormComponent,
+    value: String
+  )(implicit
+    messages: Messages,
+    sse: SmartStringEvaluator
+  ): ValidatedType[Unit] =
+    (fieldValue.errorShortNameStart, fieldValue.errorExample) match {
```
josef-vlach, line 779: All matching on `fieldValue.errorExample` are just rebinding the value, rather than doing any matching, can we drop that?

app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
```diff
@@ -654,6 +673,169 @@ object ComponentValidator {
...
+    fieldValue: FormComponent,
+    value: String,
+    maxWhole: Int
+  )(implicit
+    messages: Messages,
+    sse: SmartStringEvaluator
+  ): ValidatedType[Unit] =
+    (fieldValue.errorShortNameStart, fieldValue.errorExample) match {
+      case (None, maybeErrorExample) =>
+        val errorExample = maybeErrorExample.getOrElse(SmartString.blank)
```
josef-vlach, line 745:
```
val errorExample = maybeErrorExample.getOrElse(SmartString.blank)
```
This is duplicated 9 times, can we do something to reduce it to have this logic only once?

josef-vlach (changes requested):
There seems to be a lot of space for reducing unnecessary duplications in this PR

app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
```diff
@@ -654,6 +673,165 @@ object ComponentValidator {
...
+    fieldValue: FormComponent,
+    value: String,
+    isPositive: Boolean,
+    isWhole: Boolean
+  )(implicit
+    messages: Messages,
+    sse: SmartStringEvaluator
+  ): ValidatedType[Unit] = {
+    val maxWhole = ValidationValues.sterlingLength
+    val errorExample = fieldValue.errorExample.getOrElse(SmartString.blank)
```
josef-vlach, line 686: I think if would be worth to make `errorExampleOrBlank` property of `FormComponent` and use that instead, rather than passing it around.

app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
```diff
@@ -654,6 +673,165 @@ object ComponentValidator {
...
+          fieldValue,
+          genericPositiveSterlingErrorPositivePattern,
+          Some(
+            List(
+              errorShortNameStart.value(),
+              errorExample.value()
+            )
+          )
+        )
+    }
```
josef-vlach, line 834:
```suggestion
validationFailure(
      fieldValue,
      genericPositiveSterlingErrorPositivePattern,
      Some(
        List(
          fieldValue.errorShortNameStart match {
            case None                      => SmartString.blank.trasform(_ => "Amount", _ => "swm").value()
            case Some(errorShortNameStart) => errorShortNameStart.value()
          },
          errorExample.value()
        )
      )
    )
```

Why not move pattern matching where it really make a difference? 

Please inspect other new methods if similar improvement is possible.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/validation/ComponentValidator.scala
lxol: To be honest, I'm uncertain about that. So far, I've come across situations where, for instance, if errorShortNameStart is defined, we use one message template, and if it's not, we use a different one. I can envision a similar scenario for errorExample. If we set it to be blank by default, we won't distinguish it from the cases when a designer made it explicitly to be blank.

josef-vlach: To clarify I'm talking about adding this to the `FormComponent`
```
val errorExampleOrBlank = errorExample.getOrElse(SmartString.blank) 
```
It is just a convenience value to be used when needed.

josef-vlach (approved):

PR #1882, opened 2023-06-30
GFORMS-2214 - Add support for day(date) and month(date) in calendarDate

DursunGundogan:


app/uk/gov/hmrc/gform/eval/DateExprEval.scala
```diff
@@ -83,6 +83,13 @@ object DateExprEval {
                     _ => fromValue(evaluationContext, formComponentId),
                     localDate => DateResult(localDate)
                   )
+              case (None, Some(VariadicValue.One(m)), Some(VariadicValue.One(d))) =>
+                //Year is empty for calendar date component, so retrieves the current year to create a LocalDate.
+                Try(LocalDate.of(LocalDate.now().getYear, m.toInt, d.toInt))
```
josef-vlach, line 88: This doesn't behave correctly for leap years.
```
"includeIf": "${day(accountingDate) = 29}",
```
This includeIf is `false` for 29.02. and `true` for 29.01. It should be `true` for both cases.

josef-vlach (changes requested):

josef-vlach (approved):

PR #1905, opened 2023-07-21
GFORMS-2266 - Ensure that a user accessing a form with 'hmrcVerified'…

DursunGundogan:
… and CL 50 is sent to the IV journey

### DIFF
--- a/app/uk/gov/hmrc/gform/auth/AuthService.scala
+++ b/app/uk/gov/hmrc/gform/auth/AuthService.scala
@@ -22,7 +22,6 @@ import cats.implicits._
 import org.slf4j.LoggerFactory
 import play.api.libs.json.{ JsError, JsSuccess, Json }
 import play.api.mvc.{ AnyContent, Cookie, Request }
-import uk.gov.hmrc.auth.core.EnrolmentIdentifier
 import uk.gov.hmrc.auth.core.authorise._
 import uk.gov.hmrc.auth.core.{ AuthConnector => _, _ }
 import uk.gov.hmrc.gform.auth.models._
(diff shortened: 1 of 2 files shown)


app/uk/gov/hmrc/gform/auth/AuthService.scala
```diff
@@ -330,27 +328,22 @@ class AuthService(
...
-        val maybeEnrolmentId: Option[EnrolmentIdentifier] =
-          enrolments.getEnrolment(irsa.name).flatMap(_.getIdentifier(irsa.id))
-
-        maybeEnrolmentId.fold[AuthResult](
-          AuthRedirect(
-            gform.routes.IdentityVerificationController.enrolmentsNeeded(formTemplate._id).url
-          )
-        )(_ => authResult)
-
+      case AuthSuccessful(AuthenticatedRetrievals(_, _, AffinityGroup.Organisation, _, _, _, _), _) =>
```
lxol, line 340: Now we just block users that have organization affinity group. It was a bit more complex behavior here https://jira.tools.tax.service.gov.uk/browse/GFORMS-819 
Is that requirements no longer needed?

conf/application-json-logger.xml
```diff
@@ -5,11 +5,11 @@
         <encoder class="uk.gov.hmrc.play.logging.JsonEncoder"/>
     </appender>
 
-    <logger name="application" level="WARN"/>
+    <logger name="application" level="INFO"/>
     <logger name="destinations" level="WARN"/>
     <logger name="uk.gov" level="WARN"/>
 
-    <root level="WARN">
+    <root level="INFO">
```
lxol, line 12: I don't think it is a good idea to raise the log level to INFO in root.
I found for example that we log JWTPayload in info.
case class JwtPayload(
  sub: String,
  username: String,
  exp: Long,
  iss: String
)
don't you want to enable the info only for a particular package?

PR #1920, opened 2023-08-11
GFORMS-2236: Allow for a Welsh version of a download template

lxol:


app/uk/gov/hmrc/gform/gform/DownloadController.scala
```diff
@@ -53,9 +53,15 @@ class DownloadController(
     ) { _ => _ => _ => _ => formModelOptics =>
       val formModel = formModelOptics.formModelRenderPageOptics.formModel
       val allExprs = formModel.brackets.toBracketsPlains.toList.flatMap(_.allExprs(formModel))
+      val existInternalLink = allExprs.exists {
+        case LinkCtx(InternalLink.Download(fileName))               => true
+        case IfElse(_, LinkCtx(InternalLink.Download(fileName)), _) => true
+        case IfElse(_, _, LinkCtx(InternalLink.Download(fileName))) => true
+        case _                                                      => false
+      }
```
josef-vlach, line 61: It feels like `allExprs` should contain download link. So there must be issue elsewhere. 

Trying deep pattern matching is not a correct fix in this case.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/gform/DownloadController.scala
josef-vlach: All I'm saying is we have an error in the logic by which we get `allExprs`.

lxol:
no longer needed. Ticket is closed.

PR #1926, opened 2023-08-22
GFORMS-2288: When an ATL item is not eligible for inclusion then add …

lxol:
…feature to remove once continue is clicked


josef-vlach (approved):
Commented in the form of code contribution, I hope you don't mind.

PR #1933, opened 2023-09-04
GFORMS-2301: Initialize the registers before running report so that c…

lxol:
…orrect errors reported for lookup components


app/uk/gov/hmrc/gform/testonly/TestOnlyErrorMessageController.scala
```diff
@@ -322,7 +325,7 @@ class TestOnlyErrorMessageController(
           formComponent,
           cache.toCacheData,
           EnvelopeWithMapping.empty,
-          new LookupRegistry(Map()),
+          testLookupData,
```
josef-vlach, line 328: Why are you mocking this? Why not to pass in full `lookupRegistry` ?

josef-vlach (approved):

PR #1938, opened 2023-09-06
GFORMS-2313 - Investigate 2 accessibility issues in latest report

DursunGundogan:


app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -1140,7 +1140,7 @@ class SectionRenderingService(
           case Some(upscanData) =>
             val fileUploadName = "file"
             val otherAttributes = Map("form" -> upscanData.formMetaData.htmlId)
-            val attributes = Map("upscan" -> "upscan") ++ otherAttributes
+            val attributes = Map("data-upscan" -> "upscan") ++ otherAttributes
```
josef-vlach, line 1143: Is this attribute actually used? I don't see any js changes, so why we would need it?

DursunGundogan: This attribute is not actually used in js. I will remove it.

josef-vlach (approved):

PR #1963, opened 2023-10-02
GFORMS-2337: Make form builder extension work on pages with hidden fields when adding a note

lxol:


app/uk/gov/hmrc/gform/builder/BuilderController.scala
```diff
@@ -78,7 +78,7 @@ class BuilderController(
         gformConnector.getFormTemplateRaw(formTemplateId).map { formTemplateRaw =>
           val jsonString: String = PlayJson.stringify(formTemplateRaw)
           val maybeCirceJson: Either[ParsingFailure, Json] =
-            io.circe.parser.parse(jsonString)
+            io.circe.parser.parse(jsonString).map(j => removeObjectsByIds(j, hiddenFormComponentIds(formModelOptics)))
```
josef-vlach, line 81: Modifying original json is not composable. In the future we may want to know what is hidden on the frontend etc.

Feel free to return to the frontend result of `hiddenFormComponentIds` though. Frontend can then do the filtering itself.

josef-vlach (changes requested):

josef-vlach (approved):

PR #1976, opened 2023-10-18
GFORMS-2262 - Allow a dataRetrieve section to be configured at form l…

DursunGundogan:
…evel so data available in 1st section

### DIFF
--- a/app/uk/gov/hmrc/gform/models/DataRetrieveAll.scala
+++ b/app/uk/gov/hmrc/gform/models/DataRetrieveAll.scala
@@ -16,7 +16,6 @@
 
 package uk.gov.hmrc.gform.models
 
-import cats.data.NonEmptyList
 import cats.syntax.eq._
 import uk.gov.hmrc.gform.sharedmodel.{ DataRetrieve, DataRetrieveId }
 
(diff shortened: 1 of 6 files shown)


app/uk/gov/hmrc/gform/models/DataRetrieveAll.scala
```diff
@@ -30,9 +31,10 @@ final case class DataRetrieveAll(lookup: Map[DataRetrieveId, DataRetrieve]) exte
 object DataRetrieveAll {
   val empty: DataRetrieveAll = DataRetrieveAll(Map.empty)
 
-  def from[A <: PageMode](formModel: FormModel[A]): DataRetrieveAll =
+  def from[A <: PageMode](formModel: FormModel[A], dataRetrieve: Option[NonEmptyList[DataRetrieve]]): DataRetrieveAll =
```
lxol, line 34: why do you need to add additional argument here? dataRetrieve is in formModel already `formModel.dataRetrieve`

lxol: My IDE shows that the only usage of DataRetrieveAll is from here:
https://github.com/hmrc/gform-frontend/blob/2b74705f47cd691cfd851fcc72a83e59bdddf739/app/uk/gov/hmrc/gform/models/FormModel.scala#L155
it takes `this` and `dataRetrieve`, from from what I understand `this.dataRetrieve` is the same as `dataRetrieve` in this context.

DursunGundogan: Fixed

lxol (approved):

PR #1984, opened 2023-10-27
GFORMS-2384: Add ability to include the Direct Debit image on a page

lxol:


app/uk/gov/hmrc/gform/gform/ImageController.scala
```diff
@@ -0,0 +1,74 @@
...
+        if (file.exists()) {
+          val path = Paths.get(file.getAbsolutePath)
+          val image = java.nio.file.Files.readAllBytes(path)
+          Future.successful(
+            Ok(image).as("image/png")
+          )
+        } else {
+          Future.failed(new NotFoundException(s"File $fileName does not exist"))
+        }
+      }
```
josef-vlach, line 72: Let the play framework do the heavy lifting

```suggestion
     Future.successful {
        if (!allExprs.contains(LinkCtx(InternalLink.Image(fileName)))) {
          NotFound(
            s"link.image.$fileName expr does not exist in $formTemplateId form"
          )
        } else {
          environment.getExistingFile(s"conf/resources/$fileName") match {
            case Some(file) => Ok.sendFile(file)
            case None       => NotFound(s"File $fileName does not exist")
          }
        }
      }
```

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/gform/ImageController.scala
lxol: done suggested changes :+1:

josef-vlach (approved):

PR #1987, opened 2023-11-02
GFORMS-2343 - Sprint 104 release activities

DursunGundogan:
- Fixed top-level `data-retrieve` issue, when login as Agent

### DIFF
--- a/app/uk/gov/hmrc/gform/gform/NewFormController.scala
+++ b/app/uk/gov/hmrc/gform/gform/NewFormController.scala
@@ -372,16 +372,20 @@ class NewFormController(
 
     auth.authWithoutRetrievingForm(formTemplateId, OperationWithoutForm.EditForm) {
       implicit request => implicit lang => cache =>
-        (cache.formTemplate.draftRetrievalMethod, cache.retrievals) match {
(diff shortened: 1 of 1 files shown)


app/uk/gov/hmrc/gform/gform/NewFormController.scala
```diff
@@ -470,14 +474,13 @@ class NewFormController(
     } yield r
   }
 
-  private def exitPageHandler(
-    cache: AuthCacheWithoutForm,
-    formTemplate: FormTemplate,
-    continue: (AuthCacheWithoutForm, FormTemplate) => Future[Result]
+  private def noFormEval(
```
josef-vlach, line 477: Can we name this method more meaningfully? (And not just this one particular method)

josef-vlach (changes requested):
Once upon a time `NoFormEvaluation` was just a stateless vessel for evaluating boolean expressions, but since then it have changed  and it is now stateful bit of logic, so name `NoFormEvaluation`  doesn't make any sense anymore.

We need to reevalutate the naming around `NoFormEvaluation` logic.

josef-vlach (approved):

PR #1988, opened 2023-11-06
GFORMS-1681: DAC - Non-descriptive page title - Set title and url usi…

lxol:
…ng noPIITitle only when required

### DIFF
--- a/app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
+++ b/app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
@@ -16,6 +16,7 @@
 
 package uk.gov.hmrc.gform.gform.processor
 
+import uk.gov.hmrc.gform.eval.smartstring.{ RealSmartStringEvaluatorFactory, SmartStringEvaluatorFactory }
 import cats.instances.future._
 import cats.instances.option._
 import cats.syntax.all._
@@ -47,6 +48,8 @@ import uk.gov.hmrc.http.HeaderCarrier
 import uk.gov.hmrc.gform.eval.smartstring.SmartStringEvaluationSyntax
 
 import scala.concurrent.{ ExecutionContext, Future }
+
+import play.api.i18n.Messages
 
 class FormProcessor(
   i18nSupport: I18nSupport,
@@ -392,6 +395,15 @@ class FormProcessor(
     } yield res
   }
 
-  def getSectionTitle4Ga(processData: ProcessData, sectionNumber: SectionNumber): SectionTitle4Ga =
-    sectionTitle4GaFactory(processData.formModel(sectionNumber), sectionNumber)
+  def getSectionTitle4Ga(processData: ProcessData, sectionNumber: SectionNumber)(implicit
+    messages: Messages
+  ): SectionTitle4Ga = {
+
+    val smartStringEvaluatorFactory: SmartStringEvaluatorFactory = new RealSmartStringEvaluatorFactory()
+    val formModelVisibilityOptics = processData.formModelOptics.formModelVisibilityOptics
+
+    val sse: SmartStringEvaluator =
+      smartStringEvaluatorFactory(DataOrigin.swapDataOrigin(formModelVisibilityOptics))(messages, LangADT.En)
+    sectionTitle4GaFactory(processData.formModel(sectionNumber), sectionNumber)(sse)
+  }
 }
--- a/app/uk/gov/hmrc/gform/sharedmodel/formtemplate/SectionTitle4Ga.scala
+++ b/app/uk/gov/hmrc/gform/sharedmodel/formtemplate/SectionTitle4Ga.scala
@@ -18,15 +18,17 @@ package uk.gov.hmrc.gform.sharedmodel.formtemplate
 
 import play.api.libs.json.Format
 import uk.gov.hmrc.gform.models.{ PageMode, PageModel }
-import uk.gov.hmrc.gform.sharedmodel.LangADT
+import uk.gov.hmrc.gform.eval.smartstring.SmartStringEvaluator
 
 // For Google Analytics only. Sole purpose of this value is to be present in URLs.
 case class SectionTitle4Ga(value: String) extends AnyVal
 
 object SectionTitle4Ga {
 
-  def sectionTitle4GaFactory[D <: PageMode](pageModel: PageModel[D], sectionNumber: SectionNumber) = {
-    val sectionTitle = pageModel.noPIITitle.getOrElse(pageModel.title).localised.value(LangADT.En)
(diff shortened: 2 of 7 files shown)


app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
```diff
@@ -392,6 +395,15 @@ class FormProcessor(
     } yield res
   }
 
-  def getSectionTitle4Ga(processData: ProcessData, sectionNumber: SectionNumber): SectionTitle4Ga =
-    sectionTitle4GaFactory(processData.formModel(sectionNumber), sectionNumber)
+  def getSectionTitle4Ga(processData: ProcessData, sectionNumber: SectionNumber)(implicit
```
josef-vlach, line 398: Why not just take
```
(implicit  sse: SmartStringEvaluator)
```
?

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/SectionTitle4Ga.scala
```diff
@@ -18,15 +18,17 @@ package uk.gov.hmrc.gform.sharedmodel.formtemplate
...
 case class SectionTitle4Ga(value: String) extends AnyVal
 
 object SectionTitle4Ga {
 
-  def sectionTitle4GaFactory[D <: PageMode](pageModel: PageModel[D], sectionNumber: SectionNumber) = {
-    val sectionTitle = pageModel.noPIITitle.getOrElse(pageModel.title).localised.value(LangADT.En)
+  def sectionTitle4GaFactory[D <: PageMode](pageModel: PageModel[D], sectionNumber: SectionNumber)(implicit
+    sse: SmartStringEvaluator
+  ) = {
+    val sectionTitle: String = sse(pageModel.noPIITitle.getOrElse(pageModel.title), false)
```
josef-vlach, line 31:
```suggestion
    val sectionTitle = pageModel.noPIITitle.getOrElse(pageModel.title).value()
```

Needs import 
```
import uk.gov.hmrc.gform.eval.smartstring.SmartStringEvaluationSyntax
```

app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
lxol: we can't do that here. It should be a brand new SSE with fresh data from the page not from mongo

josef-vlach: Ok, I now see why my suggestion would be an issue.

app/uk/gov/hmrc/gform/eval/smartstring/SmartStringEvaluatorFactory.scala
```diff
@@ -57,6 +57,13 @@ class RealSmartStringEvaluatorFactory() extends SmartStringEvaluatorFactory {
 
       }
 
+      override def evalEnglish(s: SmartString, markDown: Boolean): String = {
+        val englishImplicitMessages: Messages = englishMessages
+        val englishImplicitLangADT: LangADT = LangADT.En
+
+        self(formModelVisibilityOptics)(englishImplicitMessages, englishImplicitLangADT)(s, markDown)
```
josef-vlach, line 64: Would it be possible to implement `evalEnglish` without need for new `SmartStringEvaluator` instance.

Something like

```
class RealSmartStringEvaluatorFactory(englishMessages: Messages) extends SmartStringEvaluatorFactory {

  def apply(
    formModelVisibilityOptics: FormModelVisibilityOptics[DataOrigin.Mongo]
  )(implicit
    requestMessages: Messages,
    l: LangADT
  ): SmartStringEvaluator =
    new SmartStringEvaluator {

      override def apply(s: SmartString, markDown: Boolean): String =
        withMessages(s, markDown, requestMessages, l)

      override def evalEnglish(s: SmartString, markDown: Boolean): String =
        withMessages(s, markDown, englishMessages, LangADT.En)

      def withMessages(s: SmartString, markDown: Boolean, messages: Messages, lang: LangADT): String = {
        import scala.jdk.CollectionConverters._
        new MessageFormat(s.rawValue(lang))
          .format(
            s.interpolations
              .map { interpolation =>
                formatExpr(interpolation, markDown, messages)
              }
              .asJava
              .toArray
          )
      }
...
````

File: app/uk/gov/hmrc/gform/eval/smartstring/SmartStringEvaluatorFactory.scala [hunk 4]
```diff
@@ -193,6 +200,11 @@ class RealSmartStringEvaluatorFactory() extends SmartStringEvaluatorFactory {
     new SmartStringEvaluator {
       override def apply(s: SmartString, markDown: Boolean): String =
         s.rawValue(l)
+
+      override def evalEnglish(s: SmartString, markDown: Boolean): String = {
+        val englishImplicitLangADT: LangADT = LangADT.En
+        self.noForm(englishImplicitLangADT)(s, markDown)
```
josef-vlach, line 206: I think this is better:
```suggestion
        s.rawValue(LangADT.En)
```

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/eval/smartstring/SmartStringEvaluatorFactory.scala
lxol: We need to hide implicit message, and lang if we don't use new SmartStringEvaluatorFactory instance. I refactored by putting all method inside `withMessages`. It is a bit ugly now but the other solution when we pass messages and language to all methods explicitly is not better I think. WDYT

josef-vlach: Hm, I don't like "putting all method inside `withMessages`" nor is my "pass messages and language to all methods explicitly" solution any better, they both includes subtle bugs due to implicits still being in scope. Will push a better solution proposal soon.

josef-vlach: Here is better suggestion: https://github.com/hmrc/gform-frontend/blob/hide-implicit/app/uk/gov/hmrc/gform/eval/smartstring/SmartStringEvaluatorFactory.scala

lxol: yes, makes perfect sense. I updated PR with your changes. Thanks.
I still need to wire englishLanguages from ApplicationLoader. Will do that shortly.

lxol: done englishMessages rewiring form ApplicationLoader

josef-vlach (approved):

PR #1989, opened 2023-11-07
Fix upscan fileUpload in revealing choice bug

josef-vlach:


app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -1019,7 +1051,7 @@ class SectionRenderingService(
     validationResult: ValidationResult,
     obligations: Obligations,
     upscanInitiate: UpscanInitiate,
-    upscanData: Map[FormComponentId, UpscanData] = Map.empty[FormComponentId, UpscanData]
```
josef-vlach, line 1022: Default params causing bugs yet again... we should stop using them

DursunGundogan (approved):

PR #1990, opened 2023-11-07
GFORMS-2396 - Update error messages for "overseasAddress" and "postco…

DursunGundogan:
…deLookup" to use address by default

### DIFF
--- a/app/uk/gov/hmrc/gform/addresslookup/AddressLookupController.scala
+++ b/app/uk/gov/hmrc/gform/addresslookup/AddressLookupController.scala
@@ -553,7 +553,7 @@ class AddressLookupController(
   ): Action[AnyContent] =
     auth.authAndRetrieveForm[SectionSelectorType.Normal](formTemplateId, maybeAccessCode, OperationWithForm.EditForm) {
       implicit request => implicit l => cache => implicit sse => realFormModelOptics =>
-        mkSyntheticFormModelOptics(formComponentId, cache, maybeAccessCode) {
+        mkSyntheticFormModelOptics(formComponentId, cache, maybeAccessCode, realFormModelOptics) {
           syntheticFormComponent => syntheticCache => formModelOptics =>
             val cacheData = syntheticCache.toCacheData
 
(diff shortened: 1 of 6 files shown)


app/uk/gov/hmrc/gform/addresslookup/AddressLookupController.scala
```diff
@@ -629,8 +629,8 @@ class AddressLookupController(
     fastForward: List[FastForward]
   ): Action[AnyContent] =
     auth.authAndRetrieveForm[SectionSelectorType.Normal](formTemplateId, maybeAccessCode, OperationWithForm.EditForm) {
-      implicit request => implicit l => cache => implicit sse => _ =>
-        mkSyntheticFormModelOptics(formComponentId, cache, maybeAccessCode) {
+      implicit request => implicit l => cache => implicit sse => realFormModelOptics =>
+        mkSyntheticFormModelOptics(formComponentId, cache, maybeAccessCode, realFormModelOptics) {
```
lxol, line 633: Is that for the case when we enter the address manually - to make "error" properties the same as parent component properites?

DursunGundogan: Yes, when a user wants to enter an address manually, it creates a synthetic form component in this function. I think we need to copy errorShortName/errorShortNameStart from the `postcodelookup` component.

app/uk/gov/hmrc/gform/addresslookup/AddressLookupController.scala
```diff
@@ -704,7 +707,9 @@ class AddressLookupController(
       editable = true,
       submissible = true,
       derived = false,
-      errorMessage = None
+      errorMessage = None,
```
lxol, line 710: should we add errorMessage from the parent component as well?

DursunGundogan: There is no AC about `errorMessage`. I'm not sure we need to add it

lxol: @alexhurst don't you think it makes sense to mention in the ticket that if errorMessage is configured - the "manual address" page should respect that.
Also as a side note I guess the "manual address" page is not part of the error reporting. We should probably have a ticket to fix that.

lxol: @alexhurst Actually I think there is a bigger problem here. 
```      "fields": [
        {
          "type": "postcodeLookup",
          "id": "postcode1",
          "label": "Address",
          "shortName": "Business address",
          "chooseAddressLabel": "Choose your business address",
          "confirmAddressLabel": "Confirm your business address"
        }
```
The error properties here will be applies to `postcode` page. The ticket applies them to "manual address" lookup page which is not in the form but synthetically generated and there is no way to to specify them explicitly.

alexhurst: @lxol I think that is the problem with defining an errorMessage. So for this ticket defining the errorShortNames and only using them on the manual address entry page should work. I would not worry about errorMessage for now and then we can do another ticket if it is required. Or we introduce an errorMessage but that only applied to the manual address entry page.

lxol (approved):

PR #1993, opened 2023-11-15
GFORMS-2383: Improve error message displayed while clicking upload fi…

lxol:
…le button without selecting file

### DIFF
--- a/app/uk/gov/hmrc/gform/upscan/UpscanController.scala
+++ b/app/uk/gov/hmrc/gform/upscan/UpscanController.scala
@@ -25,12 +25,13 @@ import scala.concurrent.duration._
 import scala.concurrent.{ ExecutionContext, Future }
 import uk.gov.hmrc.gform.auth.models.{ OperationWithForm, OperationWithoutForm }
 import uk.gov.hmrc.gform.config.{ AppConfig, FileInfoConfig }
-import uk.gov.hmrc.gform.controllers.{ AuthenticatedRequestActions, GformFlashKeys }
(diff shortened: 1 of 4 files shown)


app/uk/gov/hmrc/gform/upscan/UpscanController.scala
```diff
@@ -196,11 +189,19 @@ class UpscanController(
 
         val flash = errorCode match {
           case Some("InvalidArgument") =>
+            val formModel = formModelOptics.formModelVisibilityOptics.formModel
+            val formComponent = formModel.fcLookup.get(formComponentId)
```
josef-vlach, line 193: This is done on 200-201, it deserves some refactoring

josef-vlach (changes requested):
There is no consistency for js and non-js users when file is not selected.

js users see error: Select a file
non-js users see error: Select a file. It must be a PDF, JPG, XLSX, ODS, DOCX, ODT, PPTX, ODP.

Also for non-js users upload of empty file case is not properly detected anymore.

lxol:
nonjs and js can't be consistently. We have different ACs for nonjs

lxol:
Empty file detection should stay the same. I will have a look tomorrow

lxol:
>Also for non-js users upload of empty file case is not properly detected anymore.
I think it has the same behavior for empty file. Looks like locally upscan returns a generic error that I changes according to ACs. I believe it will be the same in QA.

josef-vlach (approved):
Reusing is `error_action_redirect` is neat solution 👍

PR #2002, opened 2023-11-27
GFORMS-2383: Improve error message displayed while clicking upload fi…

lxol:
…le button without selecting file

bug fix rejected error message

### DIFF
--- a/app/uk/gov/hmrc/gform/upscan/UpscanController.scala
+++ b/app/uk/gov/hmrc/gform/upscan/UpscanController.scala
@@ -24,7 +24,7 @@ import play.api.mvc.{ Action, AnyContent, Flash, MessagesControllerComponents }
 import scala.concurrent.duration._
 import scala.concurrent.{ ExecutionContext, Future }
 import uk.gov.hmrc.gform.auth.models.{ OperationWithForm, OperationWithoutForm }
-import uk.gov.hmrc.gform.config.{ AppConfig, FileInfoConfig }
+import uk.gov.hmrc.gform.config.AppConfig
 import uk.gov.hmrc.gform.controllers.{ AuthCacheWithForm, AuthenticatedRequestActions, GformFlashKeys }
 import uk.gov.hmrc.gform.gform.FastForwardService
 import uk.gov.hmrc.gform.gformbackend.GformBackEndAlgebra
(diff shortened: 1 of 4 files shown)


app/uk/gov/hmrc/gform/upscan/UpscanController.scala
```diff
@@ -114,8 +114,7 @@ class UpscanController(
                       )
                     case ConfirmationFailure.UpscanFailure(FailureDetails("REJECTED", _)) =>
                       mkFlash(
-                        "file.error.type",
```
josef-vlach, line 117: `file.error.type` is message we don't like and thus should be removed from messages.... also there are still two places to refer to it in the code.

app/uk/gov/hmrc/gform/upscan/UpscanController.scala
```diff
@@ -114,8 +114,7 @@ class UpscanController(
...
-                        "",
+                        "file.error.rejected.js",
```
josef-vlach, line 117: `file.error.rejected.js` should this rather convey what the error message is about, rather then how is this message triggered. Also please drop`.js` suffix of it.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/upscan/UpscanController.scala
lxol: fixed, I left it with the same key but changed to the same content as file.error.rejected

app/uk/gov/hmrc/gform/upscan/UpscanController.scala
lxol: I renamed original file.error.rejected to file.error.invalid.argument as it was only usage was in handling S3 error "InvalidArgument" and at the same time I renamed `file.error.rejected.js` to `file.error.rejected`
By the way we don't have and ACs how to respond to the S3 errors: `InvalidArgument`, `EntityTooLarge` and `EntityTooSmall` . We send `error_action_redirect` to upscan and from what i understood from their README in case of S3 error it will be handled by the call back. Not sure if @ratheeshsm tested that scenario.

josef-vlach (approved):

PR #2003, opened 2023-11-30
GFORMS-2444 - When outputting form data for the DataStore destination…

DursunGundogan:


app/uk/gov/hmrc/gform/gform/StructuredFormDataBuilder.scala
```diff
@@ -208,6 +208,8 @@ class StructuredFormDataBuilder[D <: DataOrigin, F[_]: Monad](
     .collect {
       case fc @ IsGroup(_)              => fc.id
       case fc @ IsInformationMessage(_) => fc.id
+      case fc @ IsMiniSummaryList(_)    => fc.id
+      case fc @ IsTableComp(_)          => fc.id
```
josef-vlach, line 212: Having this change we should rename `isGroupOrInfoIds` to something like `ignoredComponentIds` ?

josef-vlach (changes requested):

josef-vlach (approved):

PR #2008, opened 2023-12-07
GFORMS-2428 - When using postcodeLookup an invalid postcode blocks us…

DursunGundogan:
…er from entering an address manually


josef-vlach:
Is this revert of https://github.com/hmrc/gform-frontend/pull/2004 ? If yes, we should do real revert.

DursunGundogan:
Yes, this is a revert of #2004.

PR #2012, opened 2023-12-11
GFORMS-2463 - Allow for a declarationSection style page (page after C…

DursunGundogan:
…YA) to be included for a task


app/uk/gov/hmrc/gform/gform/DeclarationController.scala
```diff
@@ -65,40 +66,46 @@ class DeclarationController(
   def showDeclaration(
     maybeAccessCode: Option[AccessCode],
     formTemplateId: FormTemplateId,
-    suppressErrors: SuppressErrors
+    suppressErrors: SuppressErrors,
+    maybeCoordinates: Option[Coordinates],
+    taskCompleted: Option[Boolean]
```
josef-vlach, line 71: I don't think we should be reusing functionality for `declarationSection` to process task list declaration pages, they are not real declarations. 
We should have bespoke implementation for task list declaration pages logic.

I'm fine if you want to reuse `case class DeclarationSection`, but apart from that let's not mix these two.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/tasklist/TaskListController.scala
```diff
@@ -154,4 +156,55 @@ class TaskListController(
...
+        )
+      ) { declarationPage =>
+        import i18nSupport._
+        for {
+          validationResult <- validationService
+                                .validateAllSections(
+                                  cache.toCacheData,
+                                  formModelOptics.formModelVisibilityOptics,
+                                  EnvelopeWithMapping.empty
+                                )
```
josef-vlach, line 191: Obvious question now is what exactly are we validation here?

I think it would be worth to actually try to write this as a bespoke functionality... ie a new rendering function which doesn't handle any validation result or such.

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -784,13 +783,102 @@ class SectionRenderingService(
...
     new BackLink(href = href, content = new content.Text(messages("linkText.back")))
   }
 
+  def renderTaskDeclarationSection(
+    maybeAccessCode: Option[AccessCode],
+    formTemplate: FormTemplate,
+    singleton: Singleton[DataExpanded],
+    retrievals: MaterialisedRetrievals,
+    formModelOptics: FormModelOptics[DataOrigin.Mongo],
+    coordinates: Option[Coordinates],
```
josef-vlach, line 802: This should not be `Option` here, we should wrap it in Some in kater.

josef-vlach (changes requested):
LGTM just one small comment.

josef-vlach (approved):

PR #2013, opened 2023-12-12
GFORMS-2336: Allow Acknowledgement page to be edited

lxol:


app/uk/gov/hmrc/gform/builder/BuilderController.scala
```diff
@@ -632,4 +633,147 @@ class BuilderController(
...
+      None,
+      OperationWithForm.EditFormAcknowledgement
+    ) { implicit request => _ => cache => implicit sse => formModelOptics =>
+      import i18nSupport._
+      val formTemplate = cache.formTemplate
+      val envelopeId = cache.form.envelopeId
+
+      val destinationList = formTemplate.destinations match {
+        case destinationList: DestinationList => destinationList
+        case _                                => throw new Exception("")
```
josef-vlach, line 670: Throwing exception will not display as an error on the UI.
Please use `badRequest("Please explain what went wrong here").pure[Future]`
Also change in builder is needed for this to work.
The change there will be something like
```
const error = `Error when rendering acknowledgement panel html. ${json.error}`;
```
Instead of
```
const error = `Error when rendering acknowledgement panel html. Body content: ${json}`;
```

app/uk/gov/hmrc/gform/builder/BuilderController.scala
```diff
@@ -632,4 +633,147 @@ class BuilderController(
...
+      maybeAcknowledgementSection match {
+        case Some(acknowledgementSection) =>
+          renderHtmlForAcknowledgementFormComponent(
+            acknowledgementSection,
+            formComponentId,
+            formModelOptics,
+            cache
+          )
+        case _ =>
+          BadRequest("Can not find acknowledgement section in the form").pure[Future]
```
josef-vlach, line 715: As above this needs to use `badRequest` method

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/builder/BuilderController.scala
lxol: done

josef-vlach (approved):

PR #2017, opened 2023-12-20
GFORMS-2382 - Spike:Investigate Cookie failed message authentication …

DursunGundogan:
…check

### DIFF
--- a/app/uk/gov/hmrc/gform/ApplicationLoader.scala
+++ b/app/uk/gov/hmrc/gform/ApplicationLoader.scala
@@ -272,6 +272,8 @@ class ApplicationModule(context: Context)
   )
 
   override lazy val csrfErrorHandler: CSRF.ErrorHandler = new CSRF.CSRFHttpErrorHandler(csrfHttpErrorHandler)
+  override lazy val httpRequestHandler: HttpRequestHandler = routingModule.httpRequestHandler
+
   override def router: Router = routingModule.router
 
   val customInjector: Injector =
--- a/app/uk/gov/hmrc/gform/ApplicationLoader.scala
+++ b/app/uk/gov/hmrc/gform/ApplicationLoader.scala
@@ -110,7 +110,7 @@ class ApplicationModule(context: Context)
     val httpConfiguration: HttpConfiguration =
       new HttpConfiguration.HttpConfigurationProvider(configModule.playConfiguration, configModule.environment).get
 
-    val config: SessionConfiguration = httpConfiguration.session
+    val config: SessionConfiguration = httpConfiguration.session.copy(secure = true)
     new LegacySessionCookieBaker(config, cookieSigner)
   }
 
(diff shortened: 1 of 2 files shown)


app/uk/gov/hmrc/gform/ApplicationLoader.scala
```diff
@@ -272,8 +272,6 @@ class ApplicationModule(context: Context)
   )
 
   override lazy val csrfErrorHandler: CSRF.ErrorHandler = new CSRF.CSRFHttpErrorHandler(csrfHttpErrorHandler)
-  override lazy val httpRequestHandler: HttpRequestHandler = routingModule.httpRequestHandler
```
josef-vlach, line 275: Can we remove all `CustomHttpRequestHandler` if it is not needed? 

Has this been thoroughly tested? I'm not sue why that dropping of trailing `/`  is there for.

app/uk/gov/hmrc/gform/ApplicationLoader.scala
```diff
@@ -304,4 +302,6 @@ class ApplicationModule(context: Context)
       s"Started Frontend $appName in mode ${environment.mode} at port ${application.configuration.getOptional[String]("http.port")}"
     )
   }
+
+  override def httpFilters: Seq[EssentialFilter] = frontendFiltersModule.httpFilters
```
josef-vlach, line 306: I'm juts curious... why this has been moved and changed to `def` ?

DursunGundogan: It's defined as `def` inside `BuiltInComponents`. My IDE converts it as `def`. 
`  def httpFilters: Seq[EssentialFilter]`

app/uk/gov/hmrc/gform/ApplicationLoader.scala
DursunGundogan: I reverted this change because it will cause some problems.

app/uk/gov/hmrc/gform/playcomponents/RequestHeaderService.scala
```diff
@@ -39,7 +40,7 @@ final class RequestHeaderService(
     formTemplateIdParamIndex match {
       case Some(i) if i =!= -1 =>
         val templateId = rh.uri.split("\\?")(0).split("/")(i)
-        implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequestAndSession(rh, rh.session)
+        implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(rh)
```
josef-vlach, line 43: This actually make sense. I bet session is not properly initiated at this point, so `fromRequestAndSession` ended up with failed cookie. 

Can we undo other changes and make this single line a fix?

josef-vlach (approved):

PR #2018, opened 2023-12-22
GFORMS-2463 - Allow for a declarationSection style page (page after C…

DursunGundogan:


app/uk/gov/hmrc/gform/gform/SummaryController.scala
```diff
@@ -112,7 +112,7 @@ class SummaryController(
               formModelOptics,
               None,
               None,
-              taskCompleted
+              Some(true)
```
lxol, line 115: Looks like back link will always goes to the landing page
https://github.com/hmrc/gform-frontend/blob/02499e794f83f87eb2e0e61b140ace57a2591ed0/app/uk/gov/hmrc/gform/views/summary/summary.scala.html#L49
Is that what we expect?

DursunGundogan: Because it's the main summary page for the task-list, it should back the landing page.

DursunGundogan: Originally it was `true`, but I did it as None [it means false in this `getOrElse`] wrongly.

lxol: Ok, for formSummaryPage it makes sense, if we get there, the task is definitely completed. 
One thing that I noticed is that we never check that  `taskCompleted` is None. As far as I can see `Boolean` is what we need unless I missed something. There are some controller's methods that takes it as Option[Boolean] but I can't see that they do something different on `None` either.

DursunGundogan: We check it [here](https://github.com/hmrc/gform-frontend/blob/02499e794f83f87eb2e0e61b140ace57a2591ed0/app/uk/gov/hmrc/gform/views/summary/summary.scala.html#L49)

lxol: yes that's my point we check only 2 states not 3. We don't need Option[Boolean], Boolean is enough

DursunGundogan: It's an optional parameter in the URLs. If I set it as Boolean, it will display it on every summary and declaration page URLs.

DursunGundogan: [summary routes](https://github.com/hmrc/gform-frontend/blob/main/conf/app.routes#L39-L41)  [task-declaration](https://github.com/hmrc/gform-frontend/blob/main/conf/app.routes#L106)

lxol: Ok, personally I would've made it Boolean because it either completed or not. We don't handle task in  some third state. And for the optional parameter in the url, I would've make the default value to false, which is effectively what we do now.

lxol (approved):

PR #2051, opened 2024-02-12
GFORMS-2487: fix an issue with composite auth config

lxol:


app/uk/gov/hmrc/gform/testonly/TestOnlyController.scala
```diff
@@ -528,18 +528,33 @@ class TestOnlyController(
   }
 
   def restoreAll(formTemplateId: FormTemplateId, maybeAccessCode: Option[AccessCode]) =
+    controllerComponents.actionBuilder.async { implicit request =>
+      val snapshotId = request.body.asFormUrlEncoded.get("snapshotId").head
```
josef-vlach, line 532: `.head` is unsafe, can we raise a better error when there will not be a `snapshotId`?

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/testonly/TestOnlyController.scala
lxol: oops
that was temporary for experimentation, but I forgot to change it back :(

app/uk/gov/hmrc/gform/testonly/TestOnlyController.scala
```diff
@@ -528,18 +528,37 @@ class TestOnlyController(
   }
 
   def restoreAll(formTemplateId: FormTemplateId, maybeAccessCode: Option[AccessCode]) =
+    controllerComponents.actionBuilder.async { implicit request =>
+      val maybeSnapshotId = request.body.asFormUrlEncoded.get("snapshotId").headOption
```
josef-vlach, line 532: I can see you are using this `.head` on multiple places. Would it be possible to improve it everywhere? 

Ideally we should be using `play.api.data.Form` to bind these forms (instead of low level `request.body.asFormUrlEncoded`). That should give us way to handle errors properly also.

For example edit snapshot data page is easy to crash.

josef-vlach (changes requested):

josef-vlach (approved):

PR #2057, opened 2024-02-15
GFORMS-2561 - For specimen link forms the field level includeIf state…

DursunGundogan:
…ments should be removed/disabled


josef-vlach (changes requested):
Sorry, this needs to be done on backend. Specimen is dedicated template derived from uploaded one. And as such we can remove `includeIf`s from the specimen json before we save it into mongo.

PR #2058, opened 2024-02-19
GFORMS-2480: bug fix, hidden fcid substitution.

lxol:
make hidden ATL field substitutions in the sum explicit.


josef-vlach (commented):
When we have two ATL iterations where `pageB` is hidden on second iteration then given this expression
```
Sum(Add(Constant(1),FormCtx(pageB)))
```

`main` is currently rewriting it into:
```
Add(Constant(0),Add(Constant(1),FormCtx(1_pageB)))
```

Which is obviously incorrect, because correct rewrite would be 

```
Add(Add(Constant(1),Constant(0)),Add(Constant(1),FormCtx(1_pageB)))
```

But the question is why to complicate matter with detecting what to rewrite to `Constant(0)` at this phase? We should simply expand it into:
```
Add(Add(Constant(1),FormCtx(2_pageB)), Add(Constant(1),FormCtx(1_pageB)))
```

The invisibility of `2_pageB` component will be correctlty handled on expression evaluation phase.

And then we wouldn't need this PR.

josef-vlach (approved):

PR #2079, opened 2024-03-06
GFORMS-2522: Improve on how snapshot is restored in toolbox link

lxol:


app/uk/gov/hmrc/gform/testonly/TestOnlyController.scala
```diff
@@ -907,4 +932,53 @@ class TestOnlyController(
         )
     }
 
+  private val restoreOption1 = "restoreOption1"
+  private val restoreOption2 = "restoreOption2"
```
josef-vlach, line 936: There must be some meaningful names rather then `restoreOption1` and `restoreOption2`.

josef-vlach (approved):

PR #2128, opened 2024-04-19
GFORMS-2436 - Allow for a class to be specified on header row for table

DursunGundogan:


josef-vlach (approved):

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -1763,32 +1763,18 @@ class SectionRenderingService(
...
     val headerNumericClasses = table.rows
+      .find(isVisibleValueRow)
       .map(_.values.map(v => if (isNumeric(v)) "govuk-table__header--numeric" else ""))
-      .headOption
       .getOrElse(List())
 
-    val headerClasses: List[String] = visibleHeaders.map {
-      case (header, _) if header.classes.nonEmpty => header.classes.getOrElse("")
-      case (_, index)                             => headerNumericClasses.lift(index).getOrElse("")
+    val headCells = table.header.zipWithIndex.map { case (header, index) =>
```
lxol, line 1771: Isn't the headCell is the first cell in the row? Should it also be filtered by visible as in the original implementation?

PR #2138, opened 2024-05-09
GFORMS-2661 - Result of concat expression loses formatting when place…

DursunGundogan:
…d inside if then else expression


app/uk/gov/hmrc/gform/eval/smartstring/SmartStringEvaluatorFactory.scala
```diff
@@ -82,9 +85,12 @@ private class Executor(
     new MessageFormat(s.rawValue(formModelVisibilityOptics.booleanExprResolver.resolve(_))(l))
       .format(
         s.interpolations(formModelVisibilityOptics.booleanExprResolver.resolve(_))
-          .map {
-            case concatExpr: Concat => formatConcatExpr(concatExpr, markDown)
-            case interpolation      => formatExpr(interpolation, markDown)
+          .map { expr =>
+            val substitutions = ConcatFormatSubstitutions(concat => formatConcatExpr(concat, markDown))
```
lxol, line 89: just noticed that we may move `substitutions` to upper level, no need to create it each time for each expr

lxol (approved):

PR #2147, opened 2024-05-24
GFORMS-2690 - The specimen link for "get-help-to-pay-the-right-tax-as…

DursunGundogan:
…-wealthy-individual" is not functioning correctly and is redirecting to a technical error page.

### DIFF
--- a/app/uk/gov/hmrc/gform/gform/NewFormController.scala
+++ b/app/uk/gov/hmrc/gform/gform/NewFormController.scala
@@ -492,9 +492,13 @@ class NewFormController(
     l: LangADT
   ): Future[Result] =
     for {
-      itmpRetrievals <- formTemplate.dataRetrieve.fold(Future.successful(Option.empty[ItmpRetrievals]))(_ =>
-                          auth.getItmpRetrievals(request).map(Some(_))
-                        )
+      itmpRetrievals <- if (formTemplate.isSpecimen) {
+                          Future.successful(Option.empty[ItmpRetrievals])
+                        } else {
+                          formTemplate.dataRetrieve.fold(Future.successful(Option.empty[ItmpRetrievals]))(_ =>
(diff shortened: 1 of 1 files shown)


app/uk/gov/hmrc/gform/gform/NewFormController.scala
```diff
@@ -492,9 +492,13 @@ class NewFormController(
     l: LangADT
   ): Future[Result] =
     for {
-      itmpRetrievals <- formTemplate.dataRetrieve.fold(Future.successful(Option.empty[ItmpRetrievals]))(_ =>
-                          auth.getItmpRetrievals(request).map(Some(_))
-                        )
+      itmpRetrievals <- if (formTemplate.isSpecimen) {
+                          Future.successful(Option.empty[ItmpRetrievals])
```
lxol, line 496: does it work if some fields has includeIf that depends on itmpRetrievals or is it just ignores includeIf when in specimen template mode?

DursunGundogan: When posting a template, it removes all includeIfs if the template is a specimen template. It's working as expected.

DursunGundogan: I think we don't need to call  the external APIs for the specimen template.

lxol (approved):

app/uk/gov/hmrc/gform/gform/NewFormController.scala
josef-vlach: Shouldn't we remove `dataRetrieve` from specimen instead?

josef-vlach (changes requested):
Specimen should not need any special logic.

josef-vlach (approved):
Hm, removing `dataRetrieve` is not so easy, as there may be a lot of dependencies. Let's go forward with this solution.

PR #2155, opened 2024-06-05
GFORMS-2673: Fix misalignment of choices and includeIf when amending …

lxol:
…choice includeIf to be false

### DIFF
--- a/app/uk/gov/hmrc/gform/builder/BuilderController.scala
+++ b/app/uk/gov/hmrc/gform/builder/BuilderController.scala
@@ -68,7 +68,7 @@ class BuilderController(
   }
 
   private val compatibilityVersion =
-    14; // This is managed manually. Increase it any time API used by builder extension is changed.
+    15; // This is managed manually. Increase it any time API used by builder extension is changed.
 
   // Returns section from raw json which correspond to runtime sectionNumber parameter.
   def originalSection(
@@ -152,6 +152,7 @@ class BuilderController(
               _.deepMerge(
                 Json.obj(
                   "hiddenComponentIds" := hiddenComponentIds,
+                  "hiddenChoicesLookup" := hiddenChoiceIndexes(formModelOptics, sectionNumber),
                   "version" := compatibilityVersion
                 )
               )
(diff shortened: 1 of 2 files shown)


Aashvin (approved):

Aashvin (approved):

app/uk/gov/hmrc/gform/builder/BuilderController.scala
```diff
@@ -78,88 +80,122 @@ class BuilderController(
...
-
-            bracket.fold { nonRepeatingPage =>
-              val pageModel = nonRepeatingPage.singleton
-
-              pageModel.allFormComponentIds match {
-                case Nil => None
-                case head :: _ =>
-                  val compName = head.baseComponentId
-
+        {
```
josef-vlach, line 83: Can we remove this bracket?

lxol: removed and updated pr

app/uk/gov/hmrc/gform/builder/BuilderController.scala
```diff
@@ -78,88 +80,120 @@ class BuilderController(
   ) =
     auth.authAndRetrieveForm[SectionSelectorType.Normal](formTemplateId, maybeAccessCode, OperationWithForm.EditForm) {
       implicit request => lang => cache => sse => formModelOptics =>
-        gformConnector.getFormTemplateRaw(formTemplateId).map { formTemplateRaw =>
-          val jsonString: String = PlayJson.stringify(formTemplateRaw)
-          val maybeCirceJson: Either[ParsingFailure, Json] =
-            io.circe.parser.parse(jsonString)
+        val formModel: FormModel[DataExpanded] = formModelOptics.formModelRenderPageOptics.formModel
+        val hiddenChoicesLookupFut: Future[Map[String, List[Int]]] = formModel.allFormComponents
```
josef-vlach, line 84: Would it be possible to get that hidden indexes for particular component by comparing 
```
val visibilityOptics = formModelOptics.formModelVisibilityOptics
val renderOptics = formModelOptics.formModelRenderPageOptics
```
?

I think it should be doable and advantage of it is it it can be done without need of `Future`

lxol: ok will try to do that

josef-vlach: To elaborate on the way to achieve this is that we have basically two lists of `List[Choice]` one contains all the options and seconds contains only the visible options.
From this it should be doable to get indexes of the hidden options (by for example if `includeIf` of visible component is not present in other list).

app/uk/gov/hmrc/gform/builder/BuilderController.scala
```diff
@@ -68,7 +68,26 @@ class BuilderController(
   }
 
   private val compatibilityVersion =
-    14; // This is managed manually. Increase it any time API used by builder extension is changed.
+    15; // This is managed manually. Increase it any time API used by builder extension is changed.
+
+  def hiddenChoiceIndexes(formModelOptics: FormModelOptics[DataOrigin.Mongo]): Map[String, List[Int]] = {
+    val visibilityFormComponents = formModelOptics.formModelVisibilityOptics.formModel.allFormComponents
```
josef-vlach, line 74: Can this take into consideration only relevant components? We need to process only components for given `sectionNumber: SectionNumber`

```suggestion
    val visibilityFormComponents = formModelOptics.formModelVisibilityOptics.formModel(sectionNumber).allFormComponents
```

app/uk/gov/hmrc/gform/builder/BuilderController.scala
```diff
@@ -68,7 +68,26 @@ class BuilderController(
...
+    val renderFormComponents = formModelOptics.formModelRenderPageOptics.formModel.allFormComponents
```
josef-vlach, line 75: Same as above

app/uk/gov/hmrc/gform/builder/BuilderController.scala
```diff
@@ -68,7 +68,26 @@ class BuilderController(
...
+    visibilityFormComponents
+      .zip(renderFormComponents)
```
josef-vlach, line 77: Is using `zip` safe? What if one choice component is completely hidden in visibility model?

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/builder/BuilderController.scala
lxol: yes, noticed that already and refactoring....

josef-vlach (approved):

PR #2182, opened 2024-06-26
GFORMS-2679 - gform Perfomance optimization

DursunGundogan:


josef-vlach (commented):
What purpose will this serve?

DursunGundogan:
In the performance testing, to test the endpoint `/submissions/summary/:formTemplateId/-?action=SummaryContinue&formDataFingerprint=` , we need to pass a fingerprint into the query string. It return the fingerprint of this form in test env.

josef-vlach (approved):

PR #2218, opened 2024-08-07
GFORMS-2799 - Add ability to include Account Header for forms linked to PTA

jmartinhmrc:


conf/application.conf
```diff
@@ -195,6 +195,11 @@ gform-frontend-base-url = "http://localhost" # local nginx
 agent-subscription-frontend-base-url = "http://localhost:9437"
 signout-url="https://www.tax.service.gov.uk/gg/sign-out"
 footer-accessibility-statement-url="/accessibility-statement"
+pta-home-url="/personal-account"
+pta-messages-url="/personal-account/messages"
+pta-track-url="/track"
+pta-profile-url="/personal-account/your-profile"
```
josef-vlach, line 201: Are these going to change per environment? If not then what is the point, why not just hardcoded these to make code easier to read.

app/uk/gov/hmrc/gform/views/components/HmrcCustomHeader.scala.html
```diff
@@ -0,0 +1,165 @@
+@*
```
josef-vlach, line 1: Why not to use it [`HmrcHeader`](https://github.com/hmrc/play-frontend-hmrc/blob/main/play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcHeader.scala.html) ?

josef-vlach (changes requested):

conf/application.conf
jmartinhmrc: Updated to hardcoded values

josef-vlach (dismissed):

josef-vlach (approved):

PR #2241, opened 2024-08-23
GFORMS-2779 - Add country lookup specific error message

jmartinhmrc:


app/uk/gov/hmrc/gform/validation/TextChecker.scala
```diff
@@ -207,9 +208,15 @@ object TextChecker {
 
     def lookupError: CheckProgram[Unit] = {
       val vars: List[String] = lookupLabel.label :: Nil
-      validationFailure(fieldValue, genericErrorLookup, Some(vars))
+      validationFailure(fieldValue, getLookupErrorMessageKey, Some(vars))
     }
 
+    def getLookupErrorMessageKey: String =
+      fieldValue.`type`.cast[Text].get.constraint.cast[Lookup].get.register match {
```
josef-vlach, line 215: Calling `.get` on `Option` is not safe, and is guaranteed to fail, please let's use safe alternative

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/validation/TextChecker.scala
jmartinhmrc: Ok updated! Thought it was safe in this block but read up on it after your comment

app/uk/gov/hmrc/gform/validation/TextChecker.scala
```diff
@@ -207,9 +208,19 @@ object TextChecker {
...
+    def getLookupErrorMessageKey: String = {
+      for {
+        text   <- fieldValue.`type`.cast[Text]
+        lookup <- text.constraint.cast[Lookup]
+      } yield lookup.register match {
+        case Register.Country => countryErrorLookup
+        case _                => genericErrorLookup
+      }
+    }.getOrElse(genericErrorLookup)
+
```
josef-vlach, line 223:
```suggestion
    def getLookupErrorMessageKey: String =
      fieldValue match {
        case IsText(Text(Lookup(Register.Country, _), _, _, _, _, _)) => countryErrorLookup
        case _                                                        => genericErrorLookup
      }
```

josef-vlach: Here is a suggestion with reusing custom matcher `IsText` gform provides and uses extensively. It would be nice to keep using it for consistency.

jmartinhmrc: Thanks @josef-vlach - makes perfect sense to use this existing custom matcher

josef-vlach (approved):

PR #2249, opened 2024-09-03
Change LON to be seaport

alexhurst:


Jonathan-Courage:
Moved here https://github.com/hmrc/gform-frontend/pull/2262

PR #2251, opened 2024-09-05
GFORMS-2858 - When a user returns to a form and ITMP data has changed…

DursunGundogan:
… stale data still persists in the form

### DIFF
--- a/app/uk/gov/hmrc/gform/gform/NewFormController.scala
+++ b/app/uk/gov/hmrc/gform/gform/NewFormController.scala
@@ -215,21 +215,25 @@ class NewFormController(
             },
             {
               case "continue" =>
-                cache.formTemplate.formKind.fold(_ =>
-                  fastForwardService
-                    .redirectFastForward[SectionSelectorType.Normal](
-                      cache,
-                      noAccessCode,
-                      formModelOptics,
-                      None,
-                      SuppressErrors.Yes
+                for {
+                  updatedCache <- maybeUpdateItmpCache(request, cache, formModelOptics)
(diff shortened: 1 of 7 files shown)


app/uk/gov/hmrc/gform/gform/NewFormController.scala
```diff
@@ -539,6 +543,14 @@ class NewFormController(
       }
     } yield res
 
+  private def hasItmpExpr(exprs: List[Expr]) =
+    exprs.contains(AuthCtx(AuthInfo.ItmpAddress)) ||
+      exprs.contains(AuthCtx(AuthInfo.ItmpName)) ||
+      exprs.contains(AuthCtx(AuthInfo.ItmpDateOfBirth)) ||
+      exprs.contains(AuthCtx(AuthInfo.ItmpNameLens(ItmpNameFocus.GivenName))) ||
+      exprs.contains(AuthCtx(AuthInfo.ItmpNameLens(ItmpNameFocus.MiddleName))) ||
+      exprs.contains(AuthCtx(AuthInfo.ItmpNameLens(ItmpNameFocus.FamilyName)))
```
josef-vlach, line 552: Problem with this is that it will fail to detect `AuthCtx(ItmpNameLens(ItmpNameFocus .GivenName))`  when part of bigger expression for example: 
```
${if 1 = 2 then auth.itmpName.givenName else 'NOPE'}
```
Can we fix that?

josef-vlach (changes requested):

josef-vlach (approved):

PR #2283, opened 2024-10-15
GFORMS-2929 - Show confirmation choice on the CYA page by default but have choice to hide if not required

cmanson:


project/Dependencies.scala
```diff
@@ -17,6 +17,7 @@ object Dependencies {
     "uk.gov.hmrc"                 %% "domain-play-30"             % "9.0.0",
     "org.playframework"           %% "play-json-joda"             % "3.0.2",
     "org.julienrf"                %% "play-json-derived-codecs"   % "11.0.0",
+    "ai.x"                        %% "play-json-extensions"       % "0.42.0",
```
josef-vlach, line 20: Why we need this dependency?

cmanson: So we can format case class larger than 22 arity (one new member on FormComponent takes it to 23)

josef-vlach: We have 
```
"org.julienrf"                %% "play-json-derived-codecs"   % "11.0.0",
```
for that

cmanson: Ah okay, I will change it

josef-vlach (approved):

PR #2284, opened 2024-10-16
GFORMS-2960 - Fix form builder info component

cmanson:


app/uk/gov/hmrc/gform/builder/BuilderController.scala
```diff
@@ -1039,7 +1039,7 @@ class BuilderController(
         val infoHtml = repeater.fields.fold(HtmlFormat.empty) { fieldsNel =>
           fieldsNel.toList
             .collectFirst {
-              case fc @ IsInformationMessage(info) if fc.id === formComponentId =>
+              case fc @ IsInformationMessage(info) if fc.id.baseComponentId === formComponentId.baseComponentId =>
```
josef-vlach, line 1042: How would this detect correct component if it doesn't check an index?

josef-vlach (approved):

PR #2290, opened 2024-10-18
GFORMS-2922 - Prevent exceptions in email auth session purge filter

cmanson:


josef-vlach (commented):
What are the reproduction steps for this issue?

cmanson:
> What are the reproduction steps for this issue?

I couldn't figure out how to reproduce it - which is why I wrote a test that did reproduce it, and then fixed it.

josef-vlach:
How are you sure that this test is image of real scenario? What if this issue is due to malicious user?

Looking at the fix, it is bypassing logic of handling `new-form` logic which seems risky.

If we are unable to reproduce it, why don't we gather more data about the issue by adding extra logging first?

cmanson:
> How are you sure that this test is image of real scenario? What if this issue is due to malicious user?  

Because the test calls the filter without a form template context in the request header. That's what's causing the exception.

> Looking at the fix, it is bypassing logic of handling `new-form` logic which seems risky.

It is not bypassing the new form logic... [see here?](https://github.com/hmrc/gform-frontend/pull/2290/files#diff-9b8fb1edf900358b2d00d6694bacfddf21f4acd4f1f909fb36100c0b9a920b57R64)
 
> If we are unable to reproduce it, why don't we gather more data about the issue by adding extra logging first?

I can't figure out how to get through the filter chain in a way that doesn't have the form template context in the request headers. Can you suggest what logging we should put in?

josef-vlach:
The linked change only moves the problem somewhere else, even further from the cause.

> calls the filter without a form template context in the request header

We need to understand how is this possible in practice.

My only suggestion is to log all request data we know at the time when this exception occurs. If we already know this, then we need to try harder to reproduce it.

cmanson:
> The linked change only moves the problem somewhere else, even further from the cause.
> 
> > calls the filter without a form template context in the request header
> 
> We need to understand how is this possible in practice.
> 
> My only suggestion is to log all request data we know at the time when this exception occurs. If we already know this, then we need to try harder to reproduce it.

So should we re-implement the exception, but log everything we can get out of the header and then throw it still?

josef-vlach (approved):

PR #2299, opened 2024-10-25
GFORMS-2906 - Add the ability to look up a rate from a csv

cmanson:


app/uk/gov/hmrc/gform/gform/csv/CsvTaxRateAdapter.scala
```diff
@@ -0,0 +1,73 @@
...
+    val response: Option[TaxRate] = searchResult match {
+      case Some(list) if list.size == 1 => Some(list.head)
+      case _                            => None
+    }
+
+    response.map(rate => Json.toJson(rate))
+  }
+}
+
+case class TaxRate(regime: String, code: String, rate: String, startDate: LocalDate, endDate: LocalDate)
```
josef-vlach, line 65: `rate` must be a number

josef-vlach (changes requested):

cmanson:
Hi @josef-vlach, I've addressed all PR comments now I believe.

app/uk/gov/hmrc/gform/gform/csv/CsvTaxRateAdapter.scala
```diff
@@ -33,36 +38,33 @@ class CsvTaxRateAdapter extends CsvDataRetrieveAdapter[TaxRate] {
...
-    val searchResult: Option[List[TaxRate]] = regimeRates.map(_.filter { rate =>
-      rate.code == taxRateRequest.code && (
-        rate.startDate.isEqual(requestedDate) || rate.startDate.isBefore(requestedDate)
-      ) && (
-        rate.endDate.isEqual(requestedDate) || rate.endDate.isAfter(requestedDate)
-      )
-    })
+  override def search(request: DataRetrieve.Request): Option[JsValue] =
+    Try(request.json.as[TaxRateRequest]) match {
+      case Success(taxRateRequest: TaxRateRequest) =>
```
josef-vlach, line 49: We should use `Try` only if there is no better alternative. In this case we can use `validate` [method](https://github.com/playframework/play-json/blob/main/play-json/shared/src/main/scala/play/api/libs/json/JsReadable.scala#L37)

```
request.json.validate[TaxRateRequest] match {
      case JsSuccess(taxRateRequest, _) =>
```

app/uk/gov/hmrc/gform/gform/csv/CsvTaxRateAdapter.scala
```diff
@@ -33,36 +38,33 @@ class CsvTaxRateAdapter extends CsvDataRetrieveAdapter[TaxRate] {
...
+          rate.code === taxRateRequest.code && (
+            rate.startDate.isEqual(requestedDate) || rate.startDate.isBefore(requestedDate)
+          ) && (
+            rate.endDate.isEqual(requestedDate) || rate.endDate.isAfter(requestedDate)
+          )
+        })
+        response.map(rate => Json.toJson(rate))
+      case Failure(exception) =>
+        logger.error(s"An error occurred attempting to unmarshal request: ${request.json}")
+        throw exception
```
josef-vlach, line 63: Perhaps better alternative would be to do this:
```
case JsError(e) =>
  throw new Exception(
    s"An error occurred attempting to unmarshal request: ${request.json}. Error: ${Json.stringify(JsError.toJson(e))}"
  )
```

I think explicit logging is not needed here. We can rely on default error handler logging instead.

josef-vlach (changes requested):

PR #2316, opened 2024-11-14
GFORMS-2987 - New expression to set a date from a year

cmanson:


josef-vlach (approved):
It would be nice to include working json in the final commit message for quick reference

PR #2317, opened 2024-11-18
GFORMS-2714 - Add ability to reference columns in lookup/csv - For lo…

DursunGundogan:
…okup(port) and lookup(vfrsTradeSector) plus make easier to implement for other lookups


app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -737,6 +727,53 @@ case class EvaluationResults(
...
+              .get(register)
+              .collect { case uk.gov.hmrc.gform.lookup.AjaxLookup(lookupOptions, _, _) => lookupOptions }
+              .get
+            evalLookup(lookup, column, lookupOptions, evaluationContext.lang)
+          }
+          .getOrElse(Empty)
+      }
+      .getOrElse(Empty)
+
+  private def evalLookupColumCount(
```
josef-vlach, line 752: typo

app/uk/gov/hmrc/gform/models/FormModel.scala
```diff
@@ -151,6 +152,10 @@ case class FormModel[A <: PageMode](
     fc.id.baseComponentId
   }.toSet
 
+  val lookupRegister: Map[BaseComponentId, Register] = allFormComponents.collect { case fc @ IsTextLookup() =>
+    fc.id.baseComponentId -> fc.`type`.cast[Text].map(_.constraint.cast[Lookup].map(_.register)).get.get
```
josef-vlach, line 156: This can be simplified to
```suggestion
    fc.id.baseComponentId -> lookup.register
```
If `IsTextLookup` would return `Option[Lookup]` instead of `Boolean`

josef-vlach (changes requested):

josef-vlach (approved):

PR #2326, opened 2024-11-22
GFORMS-2743 - Redundant ‘Save and continue’ button (P1)

DursunGundogan:


app/uk/gov/hmrc/gform/views/form/form_standard.scala.html
```diff
@@ -59,9 +59,6 @@
 }
 
 @body = {
-  @* this is here for the sole purpose of making it the default submit button, when fields of group exist *@
-  @displaySaveContinueButton("defaultsink", Map("tabindex" -> "-1", "aria-hidden" -> "true"))
-
```
josef-vlach, line 64: This code is there for a reason, which comment explains quite well. If we remove it we will break hitting Enter on a page with a group component will have undesired consequences.

josef-vlach (changes requested):

josef-vlach (approved):

PR #2348, opened 2024-12-12
GFORMS-2459 - Add new expression that returns the status of a task

DursunGundogan:


app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
```diff
@@ -405,6 +417,109 @@ class FormProcessor(
...
+                                } else {
+                                  TaskStatus.InProgress
+                                }
+                              coordinate -> taskStatus
+                            }
+                          }
+                          .map(notRequiredResolver.resolveNotRequired)
+                          .map(
+                            cannotStartYetResolver.resolveCannotStartYet
+                          )
```
josef-vlach, line 515: There seems to be same logic in `private def statuses` of `TaskListRenderingService.scala`.
If I understand this correctly `TaskListRenderingService.scala` should only use statuses precomputed here, right?

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
josef-vlach: At least we should have only one implementation of how task status is computed.

DursunGundogan: `TaskListRenderingService` computes the statuses when visiting landing page. But we need to compute the status of task in FormProcessor to get `In progress` status in task. I think it would be more logical to calculate and update the status of only the selected task instead of computing the status of all tasks in FormProcessor.

josef-vlach: > I think it would be more logical to calculate and update the status of only the selected task instead of computing the status of all tasks in FormProcessor.

Not evaluating all task's statuses can lead to cache of `taskIdTaskStatus` not being up to date at all time, if there is task dependency like this for example
```
{
  "title": "Check eligibility",
  "id": "applicants",
  "sections": [
    {
      "title": "Applicant",
      "fields": [
        {
          "id": "name",
          "type": "text",
          "format": "text",
          "label": "name"
        },
        {
          "id": "age",
          "includeIf": "${taskStatus(aboutYouTask) = 'Completed'}",
          "type": "text",
          "format": "text",
          "label": "Age"
        }
      ]
    }
  ]
}
```

At this point we should perhaps ask if such a case should be even allowed. Maybe we should limit where `taskStatus` function can be used to avoid these situations.

DursunGundogan: I see it's correct. It will calculate status only for tasks with taskId.

josef-vlach (approved):

PR #2351, opened 2024-12-16
GFORMS-3071 - Summary lists for miniSummaryList, summarySection and c…

DursunGundogan:
…yaPage are showing choices within <p> so the spacing is off


DursunGundogan:
Previously there was a summary-list for each summary line. So it was leaving some space. added <p> with GFORMS-1616. But this was the wrong solution.

josef-vlach (approved):

PR #2377, opened 2025-01-06
GFORMS-3123 - When there is no label on a file component no title is …

DursunGundogan:
…rendered on page when file has been uploaded

### DIFF
--- a/app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
+++ b/app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
@@ -96,6 +96,7 @@ import uk.gov.hmrc.auth.core.ConfidenceLevel
 import uk.gov.hmrc.gform.sharedmodel.formtemplate.KeyDisplayWidth.KeyDisplayWidth
 import uk.gov.hmrc.gform.models.helpers.MiniSummaryListHelper
 import uk.gov.hmrc.gform.payment.PaymentReference
+import uk.gov.hmrc.gform.sharedmodel.formtemplate.LayoutDisplayWidth.LayoutDisplayWidth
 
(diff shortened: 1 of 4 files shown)


app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -2031,6 +2032,21 @@ class SectionRenderingService(
 
     val isPageHeading = ei.formLevelHeading
 
+    val singleton = ei.singleton
+    val page = singleton.page
+
+    val formComponents = ei.formModelOptics.formModelVisibilityOptics.formModel(ei.sectionNumber).allFormComponents
+
+    val tableComps = formComponents.collect { case IsTableComp(tc) => tc }
+    val miniSummaryListComps = formComponents.collect { case IsMiniSummaryList(tc) => tc }
```
josef-vlach, line 2041: If this method is rendering file upload component, why is here logic for table component or mini summary page?

app/uk/gov/hmrc/gform/gform/SectionRenderingService.scala
```diff
@@ -2031,6 +2032,13 @@ class SectionRenderingService(
 
     val isPageHeading = ei.formLevelHeading
 
+    val singleton = ei.singleton
+    val page = singleton.page
+
+    val maybeDisplayWidth: Option[LayoutDisplayWidth] = None
```
josef-vlach, line 2038: Isn't this `None` actually a missing implementation?

DursunGundogan: Sorry, this is a missing implementation. Fixed.

PR #2384, opened 2025-01-09
GFORMS-3128 - Fix misalignment of revealing fields on radio and check…

jmartinhmrc:
…box components.

### DIFF
--- a/public/stylesheets/gform.css
+++ b/public/stylesheets/gform.css
@@ -427,3 +427,13 @@ div.gform-date-input__item {
 .hmrc-header__service-name {
     margin-right: 5px;
 }
+
+.govuk-radios > .govuk-radios__conditional {
+    margin-left: 20px;
+    padding-left: 35px;
+}
+
+.govuk-checkboxes > .govuk-checkboxes__conditional {
+    margin-left: 20px;
+    padding-left: 35px;
+}


public/stylesheets/gform.css
```diff
@@ -427,3 +427,13 @@ div.gform-date-input__item {
...
+
+.govuk-radios > .govuk-radios__conditional {
+    margin-left: 20px;
+    padding-left: 35px;
+}
+
+.govuk-checkboxes > .govuk-checkboxes__conditional {
+    margin-left: 20px;
+    padding-left: 35px;
+}
```
josef-vlach, line 439: Shouldn't these classes be defined in the govuk-frontend library?

jmartinhmrc: So this have fixed this here: https://github.com/alphagov/govuk-frontend/commit/70f74edf11b16c14761366ac8c7d596fb3c31d1d  in  govuk-frontend v5.4.1
Which is in play-frontend-hmrc v10.6.0 and we're on v9.1.0.
We could create a ticket to upgrade? Latest is v11.9.0 which is 2 major versions though

jmartinhmrc:
Closing this PR as this issue will be fixed upgrading play-frontend-hmrc

PR #2394, opened 2025-01-17
GFORMS-2733 - Download previously submitted PDF

cmanson:
To disable: "downloadPreviousSubmissionPdf" : "false"


app/uk/gov/hmrc/gform/gform/NewFormController.scala
```diff
@@ -296,26 +297,28 @@ class NewFormController(
...
-                    ) && cache.formTemplate.downloadPreviousSubmissionPdf =>
-                  true
-                case _ => false
-              }
+          maybeSubmission <- maybeForm.fold(Option.empty[Submission].pure[Future]) { form =>
+                               gformConnector.maybeSubmissionDetails(
+                                 FormIdData(cache.retrievals, cache.formTemplate._id, noAccessCode),
+                                 form.envelopeId
+                               )
+                             }
```
josef-vlach, line 305: We don't need to do this extra call if the `form` is not `Submitted`.

I think 
```
maybeSubmission <- maybeForm.filter(_.status === Submitted).fold(...
```
should do the job

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/gform/NewFormController.scala
cmanson: Yes quite right, thanks Josef. I lost this in the previous refactor.

josef-vlach (approved):

PR #2401, opened 2025-01-27
GFORMS-3106 - Fix fastforward when atl's last iteration is deleted

jmartinhmrc:


jmartinhmrc:
Fixed in this PR: https://github.com/hmrc/gform-frontend/pull/2402 with better solution

PR #2406, opened 2025-02-04
GFORMS-2802 - Allow add/remove of choices in form builder

jmartinhmrc:


builder/src/section/choice-panel.tsx
```diff
@@ -522,6 +586,20 @@ export const ChoicePanelFactory =
             />
             {generateChoiceDividerPositionInputElements(index)}
             {generateChoiceNoneChoiceInputElements(index)}
+            <button hidden={currentNumberOfChoices < 2 || !showAllFields}
+              id={"delete-button-" + index}
+              class="btn-small btn-danger"
+              onClick={(e) => { removeChoice(e, index); } }
+            >
```
josef-vlach, line 593: Placement of remove button seems non-intuitive, because it is on the same line as following option. It would be better if it was on the first line of the choice it is actually deleting.

builder/src/section/choice-panel.tsx
```diff
@@ -522,6 +586,20 @@ export const ChoicePanelFactory =
...
+                Remove choice {index + 1}
+            </button>
+            <button hidden={!showAllFields}
+              id={"add-button-" + index}
+              class="btn-small btn-success"
+              onClick={(e) => { addChoice(e, index); } }
+            >
```
josef-vlach, line 600: Could we add extra Add choice button which will add new option at the beginning of the options?

josef-vlach (changes requested):

builder/src/section/choice-panel.tsx
jmartinhmrc: Done

josef-vlach (approved):

PR #2417, opened 2025-02-20
GFORMS-3176 - Add a new audit event fileUpload to track the files that users are uploading

jmartinhmrc:


josef-vlach (changes requested):
This audit should be created/send in upscan callback which lives is gform service

jmartinhmrc:
Implementation to be done in gForm service

PR #2419, opened 2025-02-20
GFORMS-3203 - Avoid showing Access forbidden page when user uses back to return to /previous-submission page

cmanson:


app/uk/gov/hmrc/gform/gform/NewFormController.scala
```diff
@@ -461,7 +461,9 @@ class NewFormController(
                  Ok(download_then_new(frontendConfig, downloadThenNew)).pure[Future]
                } else
                  newForm(formTemplateId, cache.toAuthCacheWithoutForm, QueryParams.fromRequest(request))
-      } yield res
+      } yield res).recover { case _ =>
+        Redirect(routes.NewFormController.newOrContinue(formTemplateId))
```
josef-vlach, line 465: It is hard to reason about why this `recover` exists. Let's instead do proper check if the `submissionDetails` exists. 

There is `gformConnector.maybeSubmissionDetails` method which should be used instead of `gformConnector.submissionDetails`

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/gform/NewFormController.scala
cmanson: Hi @josef-vlach, I've made the requested changes now.

josef-vlach (approved):

DursunGundogan (approved):

PR #2428, opened 2025-03-04
GFORMS-3221 - Add new functions daysBetween(day1,day2) and weeksBetween(day1,day2) along with ability to sum

jmartinhmrc:


app/uk/gov/hmrc/gform/eval/ExpressionResult.scala
```diff
@@ -514,9 +514,12 @@ object ExpressionResult {
       s"$monthStr $year"
     }
   }
-  case class PeriodResult(value: Period) extends ExpressionResult {
-    def +(pr: PeriodResult): PeriodResult = PeriodResult(pr.value.plus(value).normalized())
+  case class PeriodResult(value: Period, days: Long = 0, weeks: Long = 0) extends ExpressionResult {
```
josef-vlach, line 517: Result of `daysBetween(day1,day2)` and `weeksBetween(day1,day2)` should be a `NumberResult` no change to `PeriodResult` should be needed. Why is it like this?

jmartinhmrc: It was to reuse doSum for summing periods in ATL but I've redone it so that summing is supported for Number Result as well. Let me know what you think!

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -1051,7 +1077,7 @@ case class EvaluationResults(
         TypeInfo(expr, StaticTypeData(ExprType.dateString, Some(Number())))
       case DataRetrieveCount(_) =>
         TypeInfo(expr, StaticTypeData(ExprType.number, Some(Number())))
-      case Period(_, _) | PeriodValue(_)            => TypeInfo(expr, StaticTypeData(ExprType.period, None))
+      case Period(_, _, _) | PeriodValue(_)         => TypeInfo(expr, StaticTypeData(ExprType.period, None))
```
josef-vlach, line 1080: This is wrong, not all periods are `ExprType.period` (due to this PR to be precise). 

Since these new functions `daysBetween` and `weeksBetween` have nothing to do with `Period` the question is why it is parser to a `Period` expr in the first place?

I think this is bringing a lot of confusion as what is actually happening. Would it be possible to introduce new `Expr` type something like 

```
sealed trait MeasurementType
object MeasurementType {
  case object Weeks extends MeasurementType
  case object Days extends MeasurementType
}

final case class Between(dateCtx1: DateCtx, dateCtx2: DateCtx, measurementType: MeasurementType) extends Expr
```

instead? That would leave `Period` expr as is.

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -880,19 +884,23 @@ case class EvaluationResults(
     implicit val m = evaluationContext.messages
 
     def loop(expr: Expr): ExpressionResult = expr match {
-      case Add(field1: Expr, field2: Expr) => loop(field1) + loop(field2)
+      case Typed(expr: Expr, _)                    => loop(expr)
+      case Add(field1: Expr, field2: Expr)         => loop(field1) + loop(field2)
+      case Subtraction(field1: Expr, field2: Expr) => loop(field1) - loop(field2)
+      case Divide(field1: Expr, field2: Expr)      => loop(field1) / loop(field2)
+      case Multiply(field1: Expr, field2: Expr)    => loop(field1) * loop(field2)
```
josef-vlach, line 891: These are not valid operations on `Period` and this should not be here. See the other comment.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
jmartinhmrc: Ok thanks - I see the issue! I will re-implement with new Expr type

josef-vlach (approved):

PR #2462, opened 2025-04-15
GFORMS-3341 - Language switch not working on download pdf/continue pages

cmanson:


josef-vlach (commented):
I'm perplexed why this machinery is needed?

Isn't this issue just a matter of missing `GET` routes for corresponding `POST` ?

josef-vlach:
To elaborate on this. The issue is that `POST` request (below) is returning the html directly rather than via redirect.

```
POST          /new-form/:formTemplateId/access-check

**Review by josef-vlach — CHANGES_REQUESTED** [review]
This is not ideal because it leaks `csrfToken` into url. Which I guess would not be deemed accessible.

See other comment for suggestions and take
```
POST          /new-form/:formTemplateId/download-decision
```

as an inspiration of how to handle POST correctly.

File: app/uk/gov/hmrc/gform/gform/LanguageSwitchController.scala [hunk 1]
```diff
@@ -59,15 +58,36 @@ class LanguageSwitchController(
...
+                        case FormAccessCode(continueOrDeletePage) => FormAccessCodeForAgents(continueOrDeletePage)
+                      }
+                      .getOrElse((formTemplate.draftRetrievalMethod, authSuccessful.retrievals) match {
+                        case (BySubmissionReference, _)                    => BySubmissionReference
+                        case (drm @ FormAccessCodeForAgents(_), IsAgent()) => drm
+                        case (other, _)                                    => other
+                      })
+                  }
+                  .pure[Future]
+              res <- (maybeDrm, maybeAccessCode) match {
```
josef-vlach, line 82: This is suspicious. Why does this function (ie. `switchToLanguageDataChange`) has some special knowledge of `accessCode` (behind passing `accessCode` for auth purposes).

I can see that this solves issue on pages where there is `accessCode` used as a form field name. 

But what about other pages where this is not the case? On these pages welsh switching is not working as expected.

In my opinion we should keep this function as is and implement welsh switching in other way.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/gform/LanguageSwitchController.scala
cmanson: So we've now got a couple of pages, where the user is logged in and the form requires an access code - but an access code isn't known yet. If we attempt to go through load cache with form before we know the access code, it redirects us to the dashboard. 

The old logic was:
* if not logged in - just switch lanugage
* if logged in - load form, data change and save, then switch language

The new logic now says
* if not logged in - just switch language
* if logged in 
 * if form requires access code and there isn't one - just switch language
 * otherwise go through standard load from, data change and save, then switch language

What scenario have I missed?

josef-vlach: Pages this PR is addressing should be using this endpoint:

```
/language/:lang
```

Rather then this one which is used when you are inside the journey itself.

```
/language/:formTemplateId/:maybeAccessCode/:lang
```

PR #2486, opened 2025-05-06
GFORMS-2386 - Email - drop "Email must not include <character set>" v…

DursunGundogan:
…alidation

### DIFF
--- a/app/uk/gov/hmrc/gform/validation/TextChecker.scala
+++ b/app/uk/gov/hmrc/gform/validation/TextChecker.scala
@@ -396,23 +396,7 @@ object TextChecker {
       ),
       nonEmptyCheck = email(fieldValue, inputText)
     )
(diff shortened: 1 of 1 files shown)


app/uk/gov/hmrc/gform/validation/TextChecker.scala
```diff
@@ -394,17 +394,7 @@ object TextChecker {
...
-          fieldValue,
-          inputText,
-          0,
-          ValidationValues.emailLimit,
-          Some(emailErrorFirstPlaceholder()),
-          true
-        ),
-        email(fieldValue, inputText)
-      ).shortCircuitProgram
+      nonEmptyCheck = email(fieldValue, inputText)
```
cmanson, line 397: Hi Dursun, should we also be removing the standard text constraint check from `emailVerifiedByCheck`?

cmanson (approved):

PR #2498, opened 2025-05-13
GFORMS-2362 - Add ability to disable all includeIf, continueIf, redirects, validIf and mandatory file uploads in a form

cmanson:


app/uk/gov/hmrc/gform/validation/ComponentsValidator.scala
```diff
@@ -178,66 +178,74 @@ class ComponentsValidator[D <: DataOrigin, F[_]: Monad](
...
-        )
-      case emailCodeFieldMatcher.EmailCodeField(emailField) =>
-        validIf(
-          new EmailFieldIdChecker[D]().runCheck(checkerDependency)
-        )
-      case Text(constraint, _, _, _, _, _, _) =>
-        validIf(
-          new TextChecker[D]().runCheck(checkerDependency)
-        )
+    val overrideValidIfs = self.cache.formTemplate.overrides.fold(false)(o => o.disableValidIfs.getOrElse(false))
```
DursunGundogan, line 181: I think we should do all in backend. `validIf(` is not a real validIf expression. We shouldn't ignore these component validators.

cmanson: Hi Dursun, yeah I discussed this with Alex and he said that if we can ignore these implicit validators as well without too much code change, then let's do it. It's so that form designers can effectively switch off all validations so they can skip through a form without having to enter all the data.

cmanson: And we've added a check to the PR job to ensure that no PR for a production form can include the overrides element - so it should never be able to leak into production.

DursunGundogan (approved):

PR #2504, opened 2025-05-16
GFORMS-3182 - Allow user and form data to be downloaded from mongo in EEITT Admin

cmanson:


app/uk/gov/hmrc/gform/gformbackend/GformBackEndService.scala
```diff
@@ -392,4 +397,16 @@ class GformBackEndService(
...
+
+  override def newForm(formTemplateId: FormTemplateId, retrievals: MaterialisedRetrievals, queryParams: QueryParams)(
+    implicit hc: HeaderCarrier
+  ): Future[FormIdData] =
+    for {
+      newFormData <-
+        gformConnector
+          .newForm(formTemplateId, UserId(retrievals), AffinityGroupUtil.fromRetrievals(retrievals), queryParams)
+      form <- gformConnector.getForm(newFormData)
+      _    <- gformConnector.upsertAuthRetrievals(AuthRetrievals(form.envelopeId, Json.toJson(retrievals)))
```
DursunGundogan, line 409: Can we do it only one rest call?

DursunGundogan: Hi Carl, I think you need an envelopeId for the second post. Can we retrieve/parse only envelope id instead of full form data?

cmanson: I could create a new backend endpoint that persists the retrievals by using FormIdData as the key. Can use that in the backed to retrieve envelopeId, then do the upsert?

DursunGundogan (approved):

PR #2505, opened 2025-05-19
GFORMS-3439 - Update SIC codes to include a flag for "AllowedRandD"

cmanson:


app/uk/gov/hmrc/gform/LookupLoader.scala
```diff
@@ -337,13 +337,16 @@ class LookupLoader {
...
+      HeaderDecoder.decoder(englishLabel, welshLabel, idColumnName, section)(
         (_: LookupLabel, _: LookupLabel, _: LookupId, _: LookupSicCodeSection)
       )
 
+    val csvWithColumns = CsvUtils.readCsvWithColumns(filename)
     def processData(columnData: ColumnData)(index: Int): (LookupDetails, LookupDetails) = {
       val (enLabel, cyLabel, id, sicCodeSection) = columnData
-      val li = SicCodeLookupInfo(id, index, sicCodeSection)
+
+      val columns = csvWithColumns.find(_(idColumnName) == id.id).get
```
DursunGundogan, line 348:
```suggestion
val columns = csvWithColumns
  .find(_(idColumnName) === id.id)
  .getOrElse(throw new NoSuchElementException(s"No record found for ID: ${id.id}"))
```

cmanson: I don't think this is necessary, as per the other 3 lookups that load all column data. The id column is hardcoded into the Scala so this will always find what it's looking for - or the functionality for a lookup would never work. If we change it for this lookup, we should change it for all, but I really don't think it's necessary.

DursunGundogan (approved):

PR #2514, opened 2025-06-03
GFORMS-3482 - Upgrade dependency bootstrap-frontend-play-30 to versio…

jmartinhmrc:
…n 9.13.0 to stop exceptions in transitive dependency (auth client).
Required use of HttpClientV2 implemented (instead of WSClient). 
Remove uk.gov.hmrc.auth.core.retrieve.Name usage (deprecated retrieval)


conf/application.conf
```diff
@@ -20,6 +20,8 @@ formMaxAttachmentSizeMB = 10
 
 restrictedFileExtensionList = ["jfif", "png"]
 
+play.modules.enabled += "uk.gov.hmrc.play.bootstrap.HttpClientV2Module"
```
josef-vlach, line 23: This line does nothing, since gform is using it's own  custom `ApplicationLoader`, so it would be better to remove it, to avoid any future confusion.

josef-vlach (changes requested):

josef-vlach (approved):

PR #2523, opened 2025-06-10
GFORMS-3098 - Make snapshotting work on agent journeys so that they a…

DursunGundogan:
…re restored


josef-vlach (changes requested):
When I save data as an individual and then try to load the data into another form as an agent, data are not transferred.

josef-vlach (approved):

PR #2557, opened 2025-07-31
GFORMS-3595 - Add expressions to help designers cater for the scenario where a tax year changes whilst a user has saved data in a form

cmanson:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/DateExpr.scala
```diff
@@ -89,13 +89,16 @@ object OffsetYMD {
 }
 
 sealed trait DateExprValue {
-  def toLocalDate =
+  def toLocalDate(maybeFormStartDate: Option[LocalDate] = None) =
     this match {
       case TodayDateExprValue                   => LocalDate.now()
       case ExactDateExprValue(year, month, day) => LocalDate.of(year, month, day)
+      case FormStartDateExprValue =>
+        maybeFormStartDate.getOrElse(throw new Exception("Cannot calculate form start date without form"))
```
josef-vlach, line 97: Rather than throwing an exception would it be possible to accept `formStartDate: LocalDate` and when we don't have an instant of a `LocalDate` to our disposal, then we would use `Instant.now()`.

Also please let's not use default value in this method signature.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/DateExpr.scala
cmanson: Alex wants it to throw an exception to generate a technical difficulties page once the fallback is removed from the reads. 

Why are default values okay in other places but not here?

cmanson: Actually that will throw an exception if used in an `if` in an exit page because it evaluates these before any form is fetched. We could default it to now() I guess - given no form is available, and hence not started.

josef-vlach: I'd like to purge default values altogether. They have been already source of couple bugs in gform. 

Basically when you call a method it should be caller responsibility to provide all params, it is error prone when this is not enforced because people do not realise they need to fulfil the contract.

josef-vlach: > Alex wants it to throw an exception to generate a technical difficulties page once the fallback is removed from the reads.

Throwing exception when this happen will happen much earlier and has no connection with this method.

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/DateExpr.scala
```diff
@@ -89,13 +89,15 @@ object OffsetYMD {
 }
 
 sealed trait DateExprValue {
-  def toLocalDate =
+  def toLocalDate(maybeFormStartDate: Option[LocalDate]): LocalDate =
```
josef-vlach, line 92: As I said previously, this should accept `formStartDate: LocalDate`. There is no reason to use `Option` for value which will always exists.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/eval/InitFormEvaluator.scala
```diff
@@ -45,6 +46,8 @@ case class InitFormEvaluator(
   ): Boolean =
     evalBooleanExpr(includeIf.booleanExpr)
 
+  lazy val now: LocalDate = LocalDate.now()
```
josef-vlach, line 49: Why is this `lazy`? 

I wouldn't be surprise to learn that `lazy` overhead is more heavyweight than simple reading of the system clock.

jmartinhmrc (approved):

josef-vlach (approved):

PR #2581, opened 2025-08-27
GFORMS-3564 - Not all data in a form is being added to CIP audit on form submission event

jmartinhmrc:


app/uk/gov/hmrc/gform/auditing/AuditService.scala
```diff
@@ -165,6 +170,184 @@ trait AuditService {
...
+        case Some(ss) =>
+          val value = sse.evalEnglish(ss, markDown = false)
+          val processedValue = if (trim) {
+            value.trim.split("\\s+").map(_.capitalize).mkString("")
+          } else {
+            value
+          }
+          Some(processedValue)
+        case None => None
+      }
```
josef-vlach, line 184: There is `.value()` method on SmartString (assuming this: `import uk.gov.hmrc.gform.eval.smartstring._`)

Why is it not being used here and there is some custom implementation instead?

jmartinhmrc: Yes it should use smartstring method for this, thanks!

PR #2582, opened 2025-08-28
GFORMS-3643 - Eliminate side effects from recalculation logic - make it pure

PavelVjalicin:
Main changes:
```
class Recalculation[F[_]: Monad, E](
  val seissEligibilityChecker: SeissEligibilityChecker[F],
  val delegatedEnrolmentCheckStatus: DelegatedEnrolmentChecker[F],
  val dbLookupCheckStatus: DbLookupChecker[F],
  val error: GraphException => E
)
Changed to:
object Recalculation

Recalculation.recalculateFormDataNew(
    data: VariadicFormData[SourceOrigin.OutOfDate],
    formModel: FormModel[Interim],
    formTemplate: FormTemplate,
    retrievals: MaterialisedRetrievals,
    thirdPartyData: ThirdPartyData,
    evaluationContext: EvaluationContext,
    messages: Messages
  )(implicit me: MonadError[F, E]): F[RecalculationResult]
  
Changed to: 
Recalculation.recalculateFormDataNew(
    data: VariadicFormData[SourceOrigin.OutOfDate],
    formModel: FormModel[Interim],
    formTemplate: FormTemplate,
    retrievals: MaterialisedRetrievals,
    evaluationContext: EvaluationContext,
    messages: Messages,
    graphDataCache: GraphDataCache
  ): RecalculationResult
```

Introduced GraphDataCacheService, used to assemble any data that is required for Recalculation (GraphDataCache), such as (InExprResolver).
GraphDataCache is created right after Authentication in  AuthenticatedRequestActions where InExprResolver is created from form template Graph and existing formData any Future calls made pre-emptively before calculations.

Any other changes were made to comply with the changes to the above interfaces.

### DIFF
--- a/app/uk/gov/hmrc/gform/sharedmodel/graph/GraphDataCacheService.scala
+++ b/app/uk/gov/hmrc/gform/sharedmodel/graph/GraphDataCacheService.scala
@@ -59,7 +59,7 @@ class GraphDataCacheService(
     val changedCacheValues = mutable.Map[(DataSource, String), Boolean]()
 
     val inExprResolverFtr = {
-      val setOfFutures = inExprs.flatMap { case In(formCtx: FormCtx, dataSource) =>
+      val setOfFutures = inExprs.collect { case In(formCtx: FormCtx, dataSource) =>
         val modelComponentId = formCtx.formComponentId.baseComponentId
         def makeCall(value: String): Future[Boolean] = dataSource match {
           case DataSource.Mongo(collectionName) => dbLookupCheckStatus(value, collectionName, hc)
@@ -81,7 +81,7 @@ class GraphDataCacheService(
           val result = booleanExprCache.get(dataSource, value).map(Future.successful).getOrElse {
 
             makeCall(value).map { result =>
-              changedCacheValues.addOne(dataSource -> value, result)
+              changedCacheValues.addOne((dataSource -> value, result))
               result
             }
           }
@@ -89,7 +89,7 @@ class GraphDataCacheService(
           val newIn = In(new FormCtx(modelComponentId.toFormComponentId), dataSource)
           result.map(newIn -> _)
         }
-      }
+      }.flatten
 
       Future
         .sequence(setOfFutures)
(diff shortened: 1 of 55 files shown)


app/uk/gov/hmrc/gform/sharedmodel/graph/GraphDataCacheService.scala
```diff
@@ -0,0 +1,134 @@
...
+              acc.addOne((key, Map.newBuilder.addAll(value).result()))
+            }
+            .result()
+        )
+
+      } else {
+        booleanExprCache
+      }
+
+      println(finalBooleanCache)
```
josef-vlach, line 129: Remove println

app/uk/gov/hmrc/gform/sharedmodel/graph/GraphDataCacheService.scala
```diff
@@ -0,0 +1,134 @@
...
+          val newIn = In(new FormCtx(modelComponentId.toFormComponentId), dataSource)
+          result.map(newIn -> _)
+        }
+      }.flatten
+
+      Future
+        .sequence(setOfFutures)
+        .map(_.toMap)
+        .map { map => (in: In) =>
+          map.getOrElse(in, false)
```
josef-vlach, line 98: Can we throw an exception in case `in` is not in cache?

With current solution not all `in` expression are evaluated so either we do something about it or we throw an exception.

Btw. this is not introduced by this PR. This problem exists in `main` too.

This is the case which should throw an exception:
```
{
  "type": "info",
  "id": "id1",
  "label": "",
  "infoType": "noformat",
  "infoText": "Your VAT number is ${if vatNumber in mongo.mtdVatNumber then 'FOO' else 'BAR'}"
}
```

(assuming `vatNumber in mongo.mtdVatNumber` is not used in `includeIf`)

Unless you want to fix it now, we will fix this properly later in other ticket.

josef-vlach (changes requested):
Nice work, just two tiny comments and one observation.

app/uk/gov/hmrc/gform/sharedmodel/graph/GraphDataCacheService.scala
```diff
@@ -0,0 +1,134 @@
...
+    booleanExprCache: BooleanExprCache
+  )(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[GraphDataCache] = {
+
+    val formModelInterim: FormModel[Interim] = fmb.expand(variadicFormData)
+    val (graph, inExprs) = DependencyGraph.toGraph(formModelInterim, AllFormTemplateExpressions(formTemplate))
+    val changedCacheValues = mutable.Map[(DataSource, String), Boolean]()
+
+    val inExprResolverFtr = {
+      val setOfFutures = inExprs.collect { case In(formCtx: FormCtx, dataSource) =>
+        val modelComponentId = formCtx.formComponentId.baseComponentId
```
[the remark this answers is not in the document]
PavelVjalicin: Resolved

josef-vlach (approved):

PR #2597, opened 2025-09-12
GFORMS-3690 - Support an expression for mandatory property so fields are conditionally optional

cmanson:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Mandatory.scala
```diff
@@ -0,0 +1,33 @@
...
+import uk.gov.hmrc.gform.eval.BooleanExprResolver
+
+final case class Mandatory(booleanExpr: BooleanExpr) extends AnyVal {
+  def eval(booleanExprResolver: BooleanExprResolver): Boolean =
+    booleanExprResolver.resolve(booleanExpr)
+}
+
+object Mandatory {
+  val TRUE: Mandatory = Mandatory(IsTrue)
+  val FALSE: Mandatory = Mandatory(IsFalse)
```
josef-vlach, line 30: Please rename this to `True` and `False`.

josef-vlach (changes requested):

PR #2699, opened 2026-05-08
GFORMS-4223 - Investigate Chrome Mobile PrintSection PDF not updating

cmanson:


Jonathan-Courage (dismissed):
Approve quick fix for testing only

app/uk/gov/hmrc/gform/eval/EvaluationResults.scala
```diff
@@ -640,6 +641,8 @@ case class EvaluationResults(
       nonEmptyExpressionResult(exprResult)
     }
 
+    def cacheBuster: Option[String] = Some(UUID.randomUUID().toString)
```
josef-vlach: Why is this `Some`? Cannot we always add this param to the link?

cmanson: Yeah okay, will change it

josef-vlach (approved):

PR #2700, opened 2026-05-11
GFORMS-4080 - add Session ID to FileUploaded audit events

jmartinhmrc:


app/uk/gov/hmrc/gform/gformbackend/GformConnector.scala
```diff
@@ -616,9 +616,13 @@ class GformConnector(httpClient: HttpClientV2, baseUrl: String) {
 
   def upscanEncrypt(formIdData: FormIdData)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Crypted] = {
     val url = s"$baseUrl/upscan/encrypt"
+    val payload = Json.obj(
+      "formIdData" -> Json.toJson(formIdData),
+      "sessionId"  -> hc.sessionId.map(_.value)
```
josef-vlach: Does session Id needs to be encrypted?

jmartinhmrc: I suppose not! Re-implemented

app/uk/gov/hmrc/gform/upscan/UpscanService.scala
```diff
@@ -82,13 +82,16 @@ class UpscanService(
...
 
     val baseUrl = appConfig.`gform-frontend-base-url`
     val codec = new URLCodec("UTF-8")
     val callback0 = codec.encode(formIdDataCrypted.value)
+    val sessionIdParam = hc.sessionId.fold("") { sid =>
+      "&sessionId=" + codec.encode(sid.value)
+    }
     val callback: String =
-      gformBaseUrl + s"/upscan/callback/${formComponentId.value}/${envelopeId.value}?formIdDataCrypted=$callback0"
+      gformBaseUrl + s"/upscan/callback/${formComponentId.value}/${envelopeId.value}?formIdDataCrypted=$callback0$sessionIdParam"
```
josef-vlach: Can we make `sessionId` mandatory param and pass an empty string as a value if we don't have sessionId (can this even happen?)

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/upscan/UpscanService.scala
jmartinhmrc: Updated. In this case it will always be populated

josef-vlach (approved):

PR #2702, opened 2026-05-18
GFORMS-3823: India FTA – API to return e-mail data held in PDS to gForm. PDS Email lookup (GET/POST)

PavelVjalicin:
I had to calculate formVisibilityOptics one extra time to get this working. Have some performance concerns, maybe something to think about after initial deployment.

Recalc is being called because populating an ATL changes the visible sections state, which changes the fast forward behaviour.


app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
```diff
@@ -458,13 +582,45 @@ class FormProcessor(
...
+        val visitedPopulateAtlPagesVisitsIndex =
+          populateAtlDataWithTemplateIndex.foldLeft(updatedVisitsIndex) { case (acc, (atlData, atlSection)) =>
+            val visitedDefault = acc.visit(SectionNumber.Classic.AddToListPage.DefaultPage(atlSection))
+            val fm = updatedFormVisibilityOptics.formModel
+            val numberOfAtlPages = fm.addToListSectionNumbers.count {
+              case classic: SectionNumber.Classic                     => classic.sectionIndex == atlSection
+              case SectionNumber.TaskList(coordinates, sectionNumber) => sectionNumber.sectionIndex == atlSection
+            }
+
+            (0 until numberOfAtlPages - 2) //-2 excludes default and add another question pages.
```
josef-vlach: There is also "cyaPage" property in ATL and it is not being skipped when ATL is populated by dataRetrieve. I think we should fix that.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
PavelVjalicin: Done.
Also, introduced data clean-up for cases where data changed.
Minor fix to empty data set retrieved, default page is no longer visited in that instance.

app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
```diff
@@ -282,6 +280,204 @@ class FormProcessor(
...
+      Some(bankAccountInsightConnector),
+      Some(gformConnector),
+      Some(fileSystemConnector),
+      Some(delegatedAgentAuthConnector)
+    )
+
+    maybeRetrieveResultF.map { r =>
+      val visOptics = visibilityOptics.addDataRetrieveResults(r.toList)
+      val populateAtlData = dataRetrieve.populateATL
+        .map(populateAtl => populateAtlWithFormData(populateAtl, visOptics, cache)(message, langADT, hc))
```
josef-vlach: `(message, langADT, hc)` these params don't need to be passed explicitly.

app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
```diff
@@ -505,18 +687,26 @@ class FormProcessor(
...
   }
 
+  def getSectionTitle4Ga(processData: ProcessData, sectionNumber: SectionNumber)(implicit
+    messages: Messages
+  ): SectionTitle4Ga =
+    getSectionTitle4Ga(processData.formModelOptics.formModelVisibilityOptics, processData.formModel, sectionNumber)
+
   private def maybeRemoveVerifiedCode(
     validatorsResult: Option[ValidatorsResult],
     formModelVisibilityOptics: FormModelVisibilityOptics[Browser],
```
josef-vlach, line 586: Why we are returning a `Seq[PopulateAtlData])` here?

I think we should introduce same restriction we have for `"callOnNoChange": true` when `"populateATL"` functionality is being used. 
This restriction as follows: fields used in params of `dataRetrieve` needs to be defined on the same page as `dataRetrieve` itself.

Restriction [code](https://github.com/hmrc/gform/blob/022e1ff27f7a7fe22fcfa28c22b482e4997fc5c7/app/uk/gov/hmrc/gform/formtemplate/FormTemplateValidator.scala#L2064) for the inspiration.

I think that with this restriction, we should be able to return only `Option[PopulateAtlData]` 🤔

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
PavelVjalicin: Made a change where if populateAtl is present, dataRetrieve field behaves as if "callOnNoChange" is true

app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
PavelVjalicin: Done

app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
```diff
@@ -458,13 +597,55 @@ class FormProcessor(
...
+          populateAtlData -> populateAtlData.fields
+            .flatMap { case FormField(mcId, value) =>
+              fm.sectionNumberLookup.get(mcId.toFormComponentId).map(_.templateSectionIndex)
+            }
+            .headOption
+            .head
+        }
+
+        val visitedPopulateAtlPagesVisitsIndex =
+          populateAtlDataWithTemplateIndex.foldLeft(updatedVisitsIndex) { case (acc, (atlData, atlSection)) =>
```
josef-vlach: This `visitedPopulateAtlPagesVisitsIndex` contains pages which are non existent.

It would be better to use real data to get existing pages only by calling `processData.formModel.addToListSectionNumbers` and filtering only section for first iteration `TemplateSectionIndex(1)`.
That would give us something like this (both cya and default page exists in this example):

```
DefaultPage(TemplateSectionIndex(1))
Page(TemplateSectionIndex(1),1,0)
TerminalPage(TemplateSectionIndex(1),1,CyaPage)
TerminalPage(TemplateSectionIndex(1),1,RepeaterPage)
```

Those data can be used to determine if default or cya pages exists.

Having inconsistent `"visitsIndex"` in user data can have unexpected consequences elsewhere in the code.

app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
```diff
@@ -458,13 +597,55 @@ class FormProcessor(
...
+            val visitedDefault = {
+              if (atlData.fields.isDefinedAt(1)) { // index 0 will the default page
```
josef-vlach: Comment 
```
// index 0 will the default page
```
seems wrong. There seems to be no relation between `atlData.fields` and existence of default page in  given atl.

See comment about using `processData.formModel.addToListSectionNumbers` to determine what pages exists.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/sharedmodel/DataRetrieve.scala
```diff
@@ -101,6 +101,15 @@ object DataRetrieveId {
     JsonUtils.valueClassFormat[DataRetrieveId, String](DataRetrieveId.apply, _.value)
 }
 
+case class PopulateATL(
+  id: String,
```
josef-vlach: Can we define this as
```suggestion
  id: AddToListId,
```
?

Note. For this we would need to port `AddToListId` to backend too.

app/uk/gov/hmrc/gform/sharedmodel/DataRetrieve.scala
```diff
@@ -101,6 +101,15 @@ object DataRetrieveId {
...
+  mapping: Map[String, Expr]
```
josef-vlach: Can we define this as
```suggestion
  mapping: Map[FormComponentId, Expr]
```
?

app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
```diff
@@ -505,18 +690,26 @@ class FormProcessor(
     )
   }
 
-  def getSectionTitle4Ga(processData: ProcessData, sectionNumber: SectionNumber)(implicit
+  def getSectionTitle4Ga(
+    formModelVisibilityOptics: FormModelVisibilityOptics[Browser],
+    formModel: FormModel[DataExpanded],
```
josef-vlach: Does this need both `formModelVisibilityOptics` and `formModel`?

app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
```diff
@@ -282,6 +280,207 @@ class FormProcessor(
...
+                throw new RuntimeException("Could not retrieve dataRetrieve data for populateATL")
+              )
+          dataRetrieveData.data match {
+            case RetrieveDataType.ObjectType(data) => Seq(data(attribute))
+            case RetrieveDataType.ListType(data)   => data.map(attrToValue => attrToValue(attribute))
+          }
+        case _ => throw new RuntimeException(s"$expr did not match for populateATL evaluation")
+      }
+
+      val atlId = FormComponentId("1_" + populateATL.id)
```
josef-vlach: This can be replaced by safer (assuming other suggestions are in place)

```suggestion
      val atlId = populateATL.id.formComponentId.withIndex(1)
```

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
PavelVjalicin: Changed how this works, I think should be good now.

app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
PavelVjalicin: Changed how this works, I think should be good now.

PavelVjalicin:
Made some FastForward changes that are potentially problematic.

app/uk/gov/hmrc/gform/sharedmodel/DataRetrieve.scala
PavelVjalicin: Done

app/uk/gov/hmrc/gform/sharedmodel/DataRetrieve.scala
PavelVjalicin: Done

app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
PavelVjalicin: Done

PavelVjalicin:
Abstracted away populateAtl code into a separate service.

app/uk/gov/hmrc/gform/gform/processor/FormProcessor.scala
PavelVjalicin: FormModel is from renderOptics. Refactored to take in formModelOptics instead.

josef-vlach (approved):

PR #2729, opened 2026-07-16
GFORMS-4548 - Postcode lookup address without lines

cmanson:


app/uk/gov/hmrc/gform/addresslookup/AddressLookupConnector.scala
```diff
@@ -63,7 +63,17 @@ object AddressLookupConnector {
                     },
                     valid => {
                       logger.info(s"Calling address lookup returned $status: Success.")
-                      ServiceResponse(NonEmptyList.fromList(valid))
+                      val resolveNoLines = valid.map { addressRecord =>
+                        if (addressRecord.address.lines.isEmpty) {
+                          addressRecord.organisation.fold(addressRecord) { organisation =>
+                            val updatedAddress = addressRecord.address.copy(lines = List(organisation))
```
josef-vlach: I think we should `capitalize` organisation. It looks like the data are all caps for some reason.

josef-vlach (approved):

PR #2744, opened 2026-08-11
GFORMS-4654: Prove gForm to HIP to Pega Create Case integration

PavelVjalicin:


PavelVjalicin:
Opened new PR for this

PR #2747, opened 2026-08-14
GFORMS-2554: added ShowAsVrn and showAsUtr

bobzoetech:


test-templates/recalculation/internal/name-of-the-serivce-url.json
jmartinhmrc: @bobzoetech can you make these test json filenames more descriptive instead of 'name-of-the-service-url.json'?

jmartinhmrc (approved):

